PR #1, opened 2015-09-14
logging: remove useless kafka.client internal tracebacks from logging

melor:
fixed to work with py2

### DIFF
--- a/kafkajournalpump/kafkajournalpump.py
+++ b/kafkajournalpump/kafkajournalpump.py
@@ -233,6 +233,7 @@ class KafkaJournalPump(ServiceDaemon):
         return True
 
     def run(self):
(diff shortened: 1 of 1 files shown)


kafkajournalpump/kafkajournalpump.py
```diff
@@ -187,7 +187,14 @@ def __init__(self, config_path):
         self.msg_buffer = MsgBuffer(cursor)
 
         if self.config.get("journal_path"):
-            self.journald_reader = Reader(path=self.config["journal_path"])
+            while True:
+                try:
+                    self.journald_reader = Reader(path=self.config["journal_path"])
+                    break
+                except FileNotFoundError as ex:
```
saaros: not going to work on python 2.

PR #20, opened 2016-03-27
Es mappings

Ormod:


journalpump/journalpump.py
```diff
@@ -249,15 +251,35 @@ def __init__(self, config, msg_buffer, stats):
         self.index_days_max = self.config.get("elasticsearch_index_days_max", 3)
         self.index_name = self.config.get("elasticsearch_index_prefix", "journalpump")
         self.es = Elasticsearch([self.elasticsearch_url], timeout=self.request_timeout)
+        self.indices = dict((key, 1) for key in self.es.indices.get_aliases().keys())
```
hnousiainen: Use set here to make this code read easier.

journalpump/journalpump.py
```diff
@@ -457,6 +483,12 @@ def run(self):
             entry = None
             try:
                 self.initialize_sender()
+                msg_buffer_length = len(self.msg_buffer)
+                if msg_buffer_length > 50000:
+                    self.log.debug("%d entries in msg buffer, slowing down a bit by sleeping",
+                                   msg_buffer_length)
+                    time.sleep(1.0)
```
hnousiainen: Is there reason why we wouldn't stop reading new entries altogether if we're unable to flush our buffer?

journalpump/journalpump.py
```diff
@@ -249,15 +251,35 @@ def __init__(self, config, msg_buffer, stats):
         self.index_days_max = self.config.get("elasticsearch_index_days_max", 3)
         self.index_name = self.config.get("elasticsearch_index_prefix", "journalpump")
         self.es = Elasticsearch([self.elasticsearch_url], timeout=self.request_timeout)
+        self.indices = {key for key in self.es.indices.get_aliases().keys()}
```
saaros: what does get_aliases() return.  a dict?  just `set(self.es.indices.get_aliases())` would be enough here - no need to explicitly iterate over keys.

journalpump/journalpump.py
```diff
@@ -239,7 +241,7 @@ def send_messages(self, message_batch):
 class ElasticsearchSender(LogSender):
     def __init__(self, config, msg_buffer, stats):
         LogSender.__init__(self, config=config, msg_buffer=msg_buffer, stats=stats,
-                           max_send_interval=config.get("max_send_interval", 0.3))
+                           max_send_interval=config.get("max_send_interval", 10.0))
```
saaros: this config option doesn't seem to be documented

PR #22, opened 2016-04-07
Max

Ormod:


journalpump/journalpump.py
```diff
@@ -501,11 +502,12 @@ def run(self):
             try:
                 self.initialize_sender()
                 msg_buffer_length = len(self.msg_buffer)
-                if msg_buffer_length > 50000:
+                if msg_buffer_length > self.config.get("msg_buffer_max_length", 50000):
                     # This makes the self.msg_buffer grow by one journal msg a second at most
```
saaros: Update comment (and log entry below) to say that we won't add more data to the buffer before something reads it.

PR #23, opened 2016-05-04
journal fd leak fix plus other fixes and refactoring

saaros:


melor:
ok for merge, waiting for test results

PR #25, opened 2016-05-11
Move away from deprecated SimpleProducer/SimpleClient

sondinhinfectious:
I've also enforced requirements.

This should work properly with with kafka-python 1.1.1 (I've run this locally on a fresh Ubuntu 16.04 install).


sondinhinfectious:
Crap. I didn't see the existing PR. Closing this as the other one works.

PR #31, opened 2017-08-24
journalpump: multiple readers and senders

melor:
NOTE: This is a backward breaking change for the json configuration file
format!

* Multiple "readers" can be operated in a single journalpump process
* Each reader has its own input settings (source file, flags, metrics
  tags, etc.)
* One reader can have zero or more "senders", i.e. the same journal
  entries can be sent to multiple destinations (e.g. log file + Kafka)

Smaller changes:

* StatsD output is buffered, making overall performance somewhat better
* More extensive progress stats are written to the output stats file
* Output stats file is disabled unless explicitly enabled in the json
  configuration file

### DIFF
--- a/journalpump/journalpump.py
+++ b/journalpump/journalpump.py
@@ -16,10 +16,8 @@ import datetime
 import json
 import kafka.common
 import logging
-import os
 import re
 import socket
-import systemd.journal
 import time
 import uuid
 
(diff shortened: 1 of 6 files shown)


saaros:
Update `NEWS` and `README.rst`

journalpump/journalpump.py
```diff
@@ -397,76 +436,203 @@ def send_messages(self, message_batch):
...
-                except FileNotFoundError as ex:
-                    self.log.warning("journal not available yet, waiting: %s: %s",
-                                     ex.__class__.__name__, ex)
-                    time.sleep(5.0)
-        else:
-            self.journald_reader = PumpReader()
+        try:
+            self.journald_reader = PumpReader(
+                files=self.config.get("journal_files"),
+                flags=self.config.get("journal_flags"),
```
saaros: I'm not sure what's the right place to do the conversion, but I'd like to see something like this:
```
flags = self.config.get("journal_flags")
if isinstance(flags, str):
    flags = reduce(
        lambda a, b: a | b,
        [getattr(systemd.journal, flag.strip()) for flag in flags.split("|")],
    )
```
so users wouldn't have to write integer flags in their configuration files.

PR #39, opened 2018-09-25
Address lintian errors and some warnings

taneli76:
This addresses all errors indicated by `lintian` and warnings other than:

```W: journalpump: binary-without-manpage usr/bin/journalpump```

I don't know if modifying the email address after-the-fact is kosher. That change can be easily dropped if so desired.

### DIFF
--- a/debian/changelog.in
+++ b/debian/changelog.in
@@ -3,7 +3,7 @@ journalpump (1.0.1) unstable; urgency=low
   * Rename to journalpump
   * Require Python 3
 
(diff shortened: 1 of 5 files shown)


taneli76:
Oh, pylint checks failed. I doubt that has anything to do with the changes here, though.

saaros:
New pylint version apparently found new problems, fixes look pretty simple.  Would you mind including them to this PR so we'll get clean tests?

saaros:
Thanks!

PR #45, opened 2019-02-05
ci: add Python 3.7 build

ojarva:


ojarva:
Requires a bit more work regarding different dependencies in newer builder.

PR #46, opened 2019-04-09
journalpump: support remote syslog

mteiste:


journalpump/journalpump.py
```diff
@@ -3,26 +3,29 @@
...
-import kafka.common
 import logging
 import re
-import requests
 import select
 import socket
-import systemd.journal
 import time
 import uuid
+from functools import reduce
```
Ormod: Please don't reorder the imports.

journalpump/rsyslog.py
```diff
@@ -0,0 +1,101 @@
...
+
+def _rfc5424(*, pri, timestamp, hostname, app_id, proc_id, msg_id, msg, sd):
+    data = f"<{pri}>1 {timestamp} {hostname} {app_id} {proc_id} {msg_id}"
+    if sd is not None:
+        data += f" [{sd}]"
+    data += f" {msg}\n"
+    return data.encode("utf-8", "replace")
+
+
+# pylint: disable=unused-argument
```
Ormod: Could name this a bit more verbose than _rfc3164, something like rfc_3164_formatter or somesuch.

test/test_journalpump.py
```diff
@@ -1,9 +1,14 @@
+import json
```
Ormod: Alphabetically this would come after the from's

journalpump/rsyslog.py
mteiste: fixed

journalpump/journalpump.py
mteiste: order restored

test/test_journalpump.py
mteiste: fixed

PR #48, opened 2019-04-16
minor improvements and fixes

mteiste:
- add reader connection state to journpump_state.json
- report current reader state to statd
- fixed rsyslog connection handling


Ormod:
Looks fine but we discussed live that it would be good to have a test for the statefile.

PR #49, opened 2019-05-20
rsyslog: add limited support for custom log format

mteiste:
- add support for custom custom log format
- fix handling of oversized messages lines
- use default severity if PRIORITY is missing
- us default facility if SYSLOG_FACILITY is a list

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -372,6 +372,23 @@ Port used by the remote syslog server.
 Facility for the syslog message if not provided by the entry being relayed.
 (see RFC5424 for list of facilities.)
 
+``default_severity`` (default ``6``)
+
+Severity for the syslog message if not provided by the entry being relayed.
+(see RFC5424 for list of priorities.)
+
+``format`` (default ``rfc5424``)
+
+Log format to use. Can be rfc3164, rfc5424 or custom.
+
+``logline`` (default ``null``)
+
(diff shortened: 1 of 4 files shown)


README.rst
```diff
@@ -372,6 +372,19 @@ Port used by the remote syslog server.
...
+Severity for the syslog message if not provided by the entry being relayed.
+(see RFC5424 for list of priorities.)
+
+``format`` (default ``rfc5424``)
+
+Log format to use. Can be rfc3164, rfc5424 or custom.
+
+``logline`` (default ``null``)
+
+Custom logline format (ignored unless format is set to custom).
```
Ormod: An example of what it looks like or some other info on how to use this would be good.

mteiste: added an example and list of supported tags

journalpump/rsyslog.py
```diff
@@ -82,9 +134,9 @@ def close(self):
     def send(self, message):
         if self.socket is None:
             self._connect()
+        self.socket.sendall(message[:self.max_msg-1])
         if len(message) >= self.max_msg:
-            message[self.max_msg - 1] = b'\n'
-        self.socket.sendall(message[:self.max_msg])
+            self.socket.sendall(b'\n')
```
Ormod: This could be optimized a bit but that can be done if it ever turns out to be an issue.

PR #50, opened 2019-06-12
use IPv4 for rsyslog connections

mteiste:


mteiste:
Not actually the issue

PR #51, opened 2019-06-24
reader: if set, only pass through messages that match searches

hnousiainen:
If searches have been defined for a given reader, only pass through messages that match those searches to senders. This allows one to output a selection of the captured messages to readers.

Previous behavior of passing full incoming stream and doing stats on searches can be accomplished by configuring two readers the searches on the first and the senders on the latter.

### DIFF
--- a/journalpump/journalpump.py
+++ b/journalpump/journalpump.py
@@ -1000,11 +1000,13 @@ class JournalObjectHandler:
             else:
                 new_entry[key.lstrip("_")] = value
 
-        self.reader.perform_searches(self.jobject)
-
         if self.jobject.cursor is None:
             self.log.debug("No more journal entries to read")
             return False
+
+        if self.reader.searches:
+            if not self.reader.perform_searches(self.jobject):
+                return True
 
         if not self.pump.check_match(new_entry):
             return True
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,9 @@
 language: python
-dist: trusty
+dist: xenial
 
 before_install:
   - "sudo apt-get update"
-  - "sudo apt-get install -y libsystemd-journal-dev libsystemd-daemon-dev libsystemd-login-dev"
+  - "sudo apt-get install -y libsystemd-dev"
 
 python:
   - "3.4"


journalpump/journalpump.py
```diff
@@ -1000,7 +1000,9 @@ def process(self):
             else:
                 new_entry[key.lstrip("_")] = value
 
-        self.reader.perform_searches(self.jobject)
+        if self.reader.searches:
```
melor: This should probably be after the `if self.jobject.cursor is None` check?

hnousiainen: I did that initially, but decided to keep the change as minimal as possible. I can change it if you'd like, I don't think it makes much of a difference in practice.

melor:
There is a bunch of gcc warnings-as-errors failures in the jobs. Travis has probably started offering more strict compilation options by detault?

```
build/temp.linux-x86_64-3.6/systemd/_daemon.o -Werror=implicit-function-declaration
    systemd/_daemon.c: In function ‘notify’:
    systemd/_daemon.c:79:28: warning: variable ‘n_fds’ set but not used [-Wunused-but-set-variable]
             int unset = false, n_fds;
                                ^
    systemd/_daemon.c: In function ‘free_names’:
    systemd/_daemon.c:215:9: error: ‘for’ loop initial declarations are only allowed in C99 mode
             for (char **n = names; *n != NULL; n++)
             ^
    systemd/_daemon.c:215:9: note: use option -std=c99 or -std=gnu99 to compile your code
    systemd/_daemon.c: In function ‘listen_fds_with_names’:
    systemd/_daemon.c:223:25: warning: unused variable ‘item’ [-Wunused-variable]
             PyObject *tpl, *item;
                             ^
    systemd/_daemon.c:223:19: warning: unused variable ‘tpl’ [-Wunused-variable]
             PyObject *tpl, *item;
                       ^
    systemd/_daemon.c:222:16: warning: unused variable ‘names’ [-Wunused-variable]
             char **names = NULL;
                    ^
    systemd/_daemon.c:220:13: warning: unused variable ‘r’ [-Wunused-variable]
             int r;
                 ^
    systemd/_daemon.c: At top level:
    systemd/_daemon.c:212:13: warning: ‘free_names’ defined but not used [-Wunused-function]
     static void free_names(char **names) {
                 ^
    error: command 'gcc' failed with exit status 1
```

hnousiainen:
Build on newer Ubuntu target is successful

PR #61, opened 2019-12-04
Record the actual error(s) causing the disconnected state

mteiste:


journalpump/journalpump.py
```diff
@@ -197,23 +198,39 @@ def get_state(self):
...
+
+        if isinstance(error, str):
+            msg = error
+        elif isinstance(error, Exception):
+            msg = self._errors.append(getattr(error, 'message', repr(error)))
+        else:
+            msg = repr(error)
+
+        if msg and msg not in self._errors:
+            self._errors.append(msg)
```
fingon: should have only one append place, and cap length of _errors (I recommend keeping most recent ones but YMMV, first one might be also interesting)

PR #67, opened 2020-05-28
aws_cloudwatch: added sender for aws cloudwatch logs

tsharju:
Adds a new sender for sending log events to AWS CloudWatch. It is possible to configure the log group and stream as well as the AWS region to be used for the sender. AWS credentials can be set in sender config. If credentials or region are left out they are configured by the `boto3` module.

### DIFF
--- a/journalpump/journalpump.py
+++ b/journalpump/journalpump.py
@@ -25,6 +25,8 @@ import socket
 import systemd.journal
 import time
 import uuid
+import botocore
+import boto3
 
 try:
     import snappy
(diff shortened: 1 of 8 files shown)


journalpump/journalpump.py
```diff
@@ -1094,6 +1183,7 @@ def _get_or_generate_json(self, field_filter, extra_field_values, data):
             timestamp = datetime.datetime.utcfromtimestamp(data["REALTIME_TIMESTAMP"])
         else:
             timestamp = datetime.datetime.utcnow()
+            data["REALTIME_TIMESTAMP"] = timestamp.timestamp()
```
mteiste, line 1186: Don't think this needs to be added.

tsharju: I agree that it is not pretty, but I'm using that in the sender since the `data["timestamp"]` contains a string formatted date and CloudWatch wants timestamp as Epoch. I don't want to parse Epoch from string timestamp so I added the `REALTIME_TIMESTAMP` to the data here. I'm not entirely sure if journald adds this to every log entry, but since there is that `if "REALTIME_TIMESTAMP" in data:` I assumed that it is not always there.

PR #72, opened 2020-08-20
tests: refactor tests and enable typechecking for core components

willrouesnel-aivenio:
Refactor the test suite to provide more visibility into test failures, as well as enabling and fixing all typechecking problems with the initial reduced typeset.


journalpump/journalpump.py
```diff
@@ -24,7 +23,7 @@
 import time
 import uuid
 
-GeoIPReader: Union[Type[GeoIPProtocol], None]
+GeoIPReader: Any
```
lionbee, line 26: Isn't it worse going to a less concrete type?

journalpump/journalpump.py
```diff
@@ -661,11 +662,12 @@ def read_single_message(self, reader):
...
         except Exception as ex:  # pylint: disable=broad-except
             self.log.exception("Unexpected exception while handling entry for %s", reader.name)
-            self.stats.unexpected_exception(ex=ex, where="mainloop", tags=self.make_tags({"app": "journalpump"}))
+            if self.stats is not None:
+                self.stats.unexpected_exception(ex=ex, where="mainloop", tags=self.make_tags({"app": "journalpump"}))
             time.sleep(0.5)
             return False
 
-    def read_all_available_messages(self, reader, hits):
+    def read_all_available_messages(self, reader, hits: MutableMapping[str, Any]):
```
lionbee, line 670: Perhaps this is a personal opinion, but I believe in all or nothing when it comes to types in method signatures

journalpump/rsyslog.py
```diff
@@ -135,6 +136,10 @@ def close(self):
     def send(self, message):
         if self.socket is None:
             self._connect()
+
+        if self.socket is None:
+            raise Exception("socket failed to connect")
```
lionbee, line 141: I'd argue that it is not the purpose of this function to check the success of `self._connect()`, which it seems to be doing with that check on the socker

journalpump/senders/kafka.py
```diff
@@ -60,6 +60,10 @@ def _init_kafka(self):
     def send_messages(self, *, messages, cursor):
         if not self.kafka_producer:
             self._init_kafka()
+
+        if self.kafka_producer is None:
+            raise Exception("kafka producer could not be initialized")
```
lionbee, line 65: This check should be in the init function.

systest/test_rsyslog.py
```diff
@@ -35,14 +36,16 @@ def __init__(self, *, workdir, logfile, port):
...
+        self.process: Optional[Popen] = None
 
         with open(self.conffile, "w") as fp:
             print(RSYSLOGD_TCP_CONF.format(logfile=logfile, port=port), file=fp)
 
     def _wait_until_running(self):
         # Wait until the rsyslogd port is available, but if it is not up in
         # five seconds assume that it has failed to start
+        if self.process is None:
+            raise RuntimeError("start must be called before invoking this function")
```
lionbee, line 48: can't we just call start instead of failing? I know it's not the intent of the api, but this seems like a hard fail that could be avoided

journalpump/senders/rsyslog.py
```diff
@@ -56,6 +56,10 @@ def _init_rsyslog_client(self):
     def send_messages(self, *, messages, cursor):
         if not self.rsyslog_client:
             self._init_rsyslog_client()
+
+        if self.rsyslog_client is None:
+            raise Exception("failed to initialize rsyslog client")
```
lionbee, line 61: I don't think it is the purpose of this function to ensure that the previous function was succesful. I'd consider moving this failure

test/journalpump/senders/test_aws_cloudwatch.py
```diff
@@ -0,0 +1,120 @@
...
+                "ResponseMetadata": {
+                    "HTTPStatusCode": 200
+                },
+                "nextSequenceToken": "token1",
+                "rejectedLogEventsInfo": {}
+            }
+        )
+        sender.send_messages(messages=[b'{"REALTIME_TIMESTAMP": 1590581737.308352}'], cursor=None)
+        assert sender._next_sequence_token == "token1"  # pylint: disable=protected-access
+        assert sender._sent_count == 1  # pylint: disable=protected-access
```
lionbee, line 45: Ideally we don't want to test internal behaviour. A refactoring of AWSCloudWatchSender will result in a refactor of the tests I am not familiar with stubber, but I would expect these tests to check the parameters passed to the API rather than the class' internal state

lionbee (dismissed):
This is a pretty large PR. Splitting this into smaller PRs that build on top of your ideas will make it easier to get reviews and make it easier to land changes on master since turn around times are shorter

test/journalpump/test_journalpump.py
```diff
@@ -473,11 +474,18 @@ def test_pump_reader(journald_server: StandaloneJournalD):
...
 
-    entries = []
+    entries: List[JournalObject] = []
+    start_time = time()
     while True:
         entry = pump_reader.get_next()
         if entry is None:
-            break
+            if {entry.entry["MESSAGE"] for entry in entries}.issuperset(set(src_logs)):
+                break
```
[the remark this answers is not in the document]
willrouesnel-aivenio: The lines we collect are always more then what we wrote because the starting up journald process logs diagnostic data (name and memory) at the start. We can't really speak to whether that will change in the future nor is it reasonable to filter it out via this interface. For the purposes of the test the only thing we want to establish is that lines that go in are seen coming out.

Hence, the collection of unique log entries should always be a superset of what we actually sent (and at least equal).

test/journalpump/test_journalpump.py
```diff
@@ -473,11 +474,18 @@ def test_pump_reader(journald_server: StandaloneJournalD):
...
+            # Timeout after 10 seconds.
+            if time() - start_time > 10:
+                break
```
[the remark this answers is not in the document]
willrouesnel-aivenio: journald starts fast, but not instantly. But our interface to it under the hood is `journalctl`. With no delays at all, this test will sometimes fail because journald hasn't flushed the logs yet. It is of course possible that `journalctl` has broken entirely in an unexpected way, in which case it'll never flush.

But in both cases the only indication we get is we call `journalctl` via this interface and it stops returning log lines (returns `None`). In the case of a catastrophic failure, a 10s delay is no big deal because the entire test suite is about to abort anyway. In practical terms it takes less then a second usually.

There's not much you can do to solve it so long as subprocesses are involved since we can't peek into their internal state and I considered, but didn't want to commit too, a pure python Journald format writer to support this test suite (not hard, not super quick to do).

journalpump/journalpump.py
willrouesnel-aivenio: Ah, this is a leftover. So originally we didn't have _any_ types. Adding a protocol for the bits of GeoIP we used was my initial solution, but since I don't have a solid grasp on what the behaviour should be it puts too tight of a bound on what is known - so I pulled it out and just set GeoIPReader to `Any` because in practice that's kind of how it's handled by the code.

journalpump/journalpump.py
willrouesnel-aivenio: There are commented out lines in `mypy.ini` which represent the settings I wanted to enable. But the practical reality of that was some of them imply large code refactors. The initial goal is to get mypy and the typechecker running and passing, so they can later be tightened (which again: requires larger code refactors to do soundly, but makes the PR even bigger).

journalpump/rsyslog.py
willrouesnel-aivenio: So this is a common structuring problem in almost all our daemons: `ProviderDaemon(someConfig) -> daemon.init() -> daemon.run() -> do_things_with_inited_resources()`

This pattern is hard to work with the typechecker because where and how things should fail is non-obvious: the answer we have is, just crash the process and let it be restarted. Which is a practical safety measure but makes it hard to modify the code because you never know if you're about to crash something somewhere else. Can self.socket be None, is that a reasonable thing (here, no, elsewhere, potentially, in both cases, nothing confirms you're being sensible or haven't missed something).

In this case, calling `self._connect()` _implies_ that self.socket is now not None ... but does nothing to assure the typechecker, or say, a programmer not looking at the rest of the code, that that's what it's actually done. So the typechecker rightly complains that self.socket can still be `Optional[socket]` after this, and in fact if someone breaks `self._connect()` in the future nothing is actually going to be able to determine it (we just crash loop).

Better - and I did start refactoring in this direction but it's a lot of changes - would be if the signature was `self._connect() -> socket.Socket`, because then the answer is explicitly "you always get a socket, or we're going to crash - don't worry about it". But it's a lot of refactoring and I don't understand the code well enough to know I wouldn't break behavior doing that.

This line could be replaced with a simple `assert self.socket is not None` to make the typechecker happy, which at runtime at least compiles down to a noop AFAIK (there's some way to do it IIRC). 

IMO the exception here is an improvement but not much of one, since if the socket actually fails to connect we kind of want to know why - DNS error (config?), did it connect previously ever etc.

journalpump/senders/kafka.py
willrouesnel-aivenio: I agree but see above: moving the check to `__init__` would give us RAII semantics which would make mypy happy which would make everything happy, but it fundamentally changes the structure of `journalpump` in a big way I wouldn't want to embark on without the initial typechecks and really solid testing.

systest/test_rsyslog.py
willrouesnel-aivenio: This is an internal function that will literally never be called anywhere else except the function right below it. This check does only 2 things: makes mypy realize self.process is not None (without changing entirely how this particular API works right now), and gives a great big warning to any future test programmer that they've messed up (but really mostly the former).

Reasonably a new PR would just move the content of `_wait_until_running` into `start` which resolves all the ambiguity entirely...but then really, this should actually be a pytest fixture which uses `yield` semantics. The issue is all of that is varying levels of surgery which doesn't get closer to having rigorous type-checks over the code.

journalpump/senders/rsyslog.py
willrouesnel-aivenio: See https://github.com/aiven/journalpump/pull/72#discussion_r475588369

The only reasonable change with the current structure is to replace it for an assert .

test/journalpump/senders/test_aws_cloudwatch.py
willrouesnel-aivenio: I did in fact refactor a substantial amount of all the senders before realizing that it wasn't reasonable to try and get them merged while I'm really just trying to get the typechecking running in CI for us. Really this test should be replaced with a `systest` which just calls the actual AWS Cloudwatch API if we're going to support it and then we can treat the unit tests like a blackbox with well-specified typechecks to keep them honest.

journalpump/journalpump.py
willrouesnel-aivenio: Oh this wasn't a leftover - there was going to be a much more complicated protocol type, but it's a lot of work for an ill-defined typespace so it was dropped in favor of Any - the code dynamically tests for features anyway.

PR #73, opened 2020-08-20
journal: add log namespace support

alexole:


alexole:
Closing for now, this requires newer python3-systemd and systemd versions, which will be available in F32.
https://app.clubhouse.io/aiven/story/17541/make-sure-tlog-auditing-doesn-t-end-up-in-journal-on-the-node-side-to-avoid-recursion

PR #78, opened 2020-10-13
kafka: check api_version

gabriel-tincu:
When enabling ZSTD, check the kafka api version , since the compression
algorithm requires a cluster version of 2.1.0 or higher

### DIFF
--- a/journalpump/senders/kafka.py
+++ b/journalpump/senders/kafka.py
@@ -37,10 +37,13 @@ class KafkaSender(LogSender):
         self.mark_disconnected()
         while self.running:
             try:
+                api_version = self.config.get("kafka_api_version", "0.9")
+                parsed = tuple(map(int, api_version.split(".")))
                 if self.kafka_producer:
                     self.kafka_producer.close()
                 # make sure the python client supports it as well
-                if zstd and "zstd" in KafkaProducer._COMPRESSORS:  # pylint: disable=protected-access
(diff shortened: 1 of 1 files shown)


hackaugusto:
This is done by `kafka-python` already:

https://github.com/dpkp/kafka-python/blob/9feeb79140ed10e3a7f2036491fc07573740c231/kafka/producer/kafka.py#L395-L396

journalpump/senders/kafka.py
```diff
@@ -37,18 +37,21 @@ def _init_kafka(self):
...
         while self.running:
             try:
+                api_version = self.config.get("kafka_api_version", "0.9")
+                parsed = tuple(map(int, api_version.split(".")))
                 if self.kafka_producer:
                     self.kafka_producer.close()
                 # make sure the python client supports it as well
-                if zstd and "zstd" in KafkaProducer._COMPRESSORS:  # pylint: disable=protected-access
+                # pylint: disable=protected-access
+                if zstd and "zstd" in KafkaProducer._COMPRESSORS and parsed >= (2, 1, 0):
```
ivanyu, line 46: Maybe, use `LooseVersion`?

ivanyu (changes requested):

fingon:
@ivanyu is this still worth doing or should we close it?

ivanyu:
I think we don't need to care about that old Kafka versions.

PR #79, opened 2020-11-03
aws_cloudwatch: fixed sender init to support log stream pagination

tsharju:
Fixes an issue where having more than page size log streams inside a log group, the log stream was not found and the sequence token needed for sending logs could not be set. Now we use pagination when listing the log streams inside the log group and paginate until the correct stream is found.

### DIFF
--- a/journalpump/senders/aws_cloudwatch.py
+++ b/journalpump/senders/aws_cloudwatch.py
@@ -38,13 +38,18 @@ class AWSCloudWatchSender(LogSender):
         except botocore.exceptions.ClientError as err:
             if err.response["Error"]["Code"] != "ResourceAlreadyExistsException":
                 raise
-        streams = self._logs.describe_log_streams(logGroupName=self.log_group).get("logStreams")
-        if streams is not None:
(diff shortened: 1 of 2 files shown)


journalpump/senders/aws_cloudwatch.py
```diff
@@ -38,13 +38,18 @@ def _init_logs(self):
         except botocore.exceptions.ClientError as err:
             if err.response["Error"]["Code"] != "ResourceAlreadyExistsException":
                 raise
```
rikonen, line 40: The entire init function should probably be wrapped in some error handler that just marks the sender as disconnected in case of any error as these too would cause entire journalpump to fail when there's some problem with CloudWatch config.

rikonen (approved):
Could further improve error handling to avoid journalpump failing completely in case of unexpected error but the particular fix done here looks good to me.

rikonen:
Looks like Travis is having problems again and tests aren't running

tsharju:
The tests seem to have gone through finally. I agree about the error handling. Somehow forgot to include it here. I'll make a separate PR for that as this is in a bit of a hurry to get in.

PR #81, opened 2020-11-11
kafkasender: by default infer the api version

hackaugusto:
Unless explicitly provided by the user, let `KafkaProducer` infer the api version. `0.9` is old enough that using this as the default can lead to errors if some modern features are used, e.g. compression with `zstd`.


ivanyu (approved):
Makes sense to me, would like to hear other opinions.

ivanyu (approved):

PR #87, opened 2021-02-16
google_cloud_logging: decode MESSAGE to json object if it's possible

axeoman:
If `MESSAGE` in journald entry is a decodable JSON - send it to GCP Logging as an object instead of string.


journalpump/senders/google_cloud_logging.py
```diff
@@ -57,6 +57,12 @@ def send_messages(self, *, messages, cursor):
             msg = json.loads(msg_str)
             timestamp = msg.pop("timestamp", None)
             journald_priority = msg.pop("PRIORITY", None)
+
+            try:
+                msg["MESSAGE"] = json.loads(msg["MESSAGE"])
+            except (json.JSONDecodeError, KeyError, TypeError):
```
tsharju, line 63: A bit of a personal preference, but maybe rather use `dict.get` for checking if `MESSAGE` exists and then decode that rather than catch the `KeyError`. I think `MESSAGE` should always contain a string so in that sense maybe better to not catch the `TypeError` as that probably is an actual error case.

tsharju: Also, using `contextlib.suppress` would be more preferable than `try: except: pass`.

tsharju (dismissed):
See the comment. Also, having a simple test for both the case where `MESSAGE` is plain string and JSON decodable string would be nice.

journalpump/senders/google_cloud_logging.py
axeoman: I had an Ask for Forgiveness approach in mind here, but yes checking for a key before looks better.
I also doubted about TypeError check here, going to remove it.

test/test_google_cloud_logging.py
```diff
@@ -0,0 +1,150 @@
...
+            stats=mock.Mock(),
+            field_filter=None,
+            config=self.CONFIG,
+            googleapiclient_request_builder=request_builder,
+        )
+        sender.send_messages(messages=[b'{"MESSAGE": "Hello"}'], cursor=None)
+        assert sender._sent_count == 1  # pylint: disable=protected-access
+
+    def test_missing_message(self):
+        """Check that MESSAGE as object is decoded was sent"""
```
tsharju, line 72: I think this test and the next one has comments the wrong way around.

tsharju (changes requested):
Looking good. Just a minor thing. See the comment.

test/test_google_cloud_logging.py
axeoman: Thanks! Fixed.

tsharju (approved):

PR #88, opened 2021-02-21
spec: disable dependency generator

axeoman:


axeoman:
pylint started to complain about `Using deprecated method wrap_socket()` so I fixed it in https://github.com/aiven/journalpump/pull/89

journalpump.spec
```diff
@@ -11,6 +11,10 @@ BuildRequires:  python3-devel, python3-pytest, python3-pylint python3-responses
 BuildArch:      noarch
 Obsoletes:      kafkajournalpump
 
+# The dependency generator uses names which are not compatible with Aiven
+# internal .spec files.
```
tsharju, line 15: I believe this is more related to Fedora package naming which I think is what we follow as well as much as possible. So maybe adjust that comment to refer to that.

tsharju (changes requested):
So the problem with the RPM dependencies was probably that the automatic dependency generator generated dependencies not compatible with Fedora naming style. So I think adding the disable for that in the .spec file is ok. What I'm concerned is that not all dependencies are  correctly listed in the RPM spec (they weren't before either), but this seems like a good opportunity to fix that. If I looked correctly the .spec file misses `Requires` entry for `python3-geoip2` and `python3-typing-extensions`.

journalpump.spec
axeoman: Thanks! Changed the sentence.

tsharju (approved):

PR #90, opened 2021-05-28
awscloudwatchsender: fix crash due to wrong credentials

jhonkola:
Journalpump would crash if `AWSCloudwatchSender` got
`AccessDeniedException` during sender init. This can
happen if there are wrong credentials in config.

All `ClientError` class AWS errors are now caught during init
and retried a few times. This will catch `AccessDeniedException`
and transient `ServiceUnavailable` errors.

`AccessDeniedException` errors are reported to `JournalReader`. 
`JournalReader` ignores senders that failed to initialize during operation.

### DIFF
--- a/journalpump/journalpump.py
+++ b/journalpump/journalpump.py
@@ -7,7 +7,7 @@ from .daemon import ServiceDaemon
 from .senders import (
     AWSCloudWatchSender, ElasticsearchSender, FileSender, GoogleCloudLoggingSender, KafkaSender, LogplexSender, RsyslogSender
 )
-from .senders.base import MAX_KAFKA_MESSAGE_SIZE, Tagged
+from .senders.base import MAX_KAFKA_MESSAGE_SIZE, SenderInitializationError, Tagged
 from .types import GeoIPProtocol
 from .util import atomic_replace_file, default_json_serialization
 from functools import reduce
@@ -103,6 +103,17 @@ class PumpReader(Reader):
 
 
 class JournalReader(Tagged):
+    # config name <--> class mapping
+    sender_classes = {
+        "elasticsearch": ElasticsearchSender,
+        "kafka": KafkaSender,
+        "logplex": LogplexSender,
+        "file": FileSender,
+        "rsyslog": RsyslogSender,
+        "aws_cloudwatch": AWSCloudWatchSender,
+        "google_cloud_logging": GoogleCloudLoggingSender,
+    }
+
     def __init__(
         self,
         *,
@@ -139,7 +150,7 @@ class JournalReader(Tagged):
         self.last_journald_create_attempt = 0
         self.running = True
         self.senders = {}
-        self._senders_initialized = False
+        self._initialized_senders = set()
         self.last_stats_send_time = time.monotonic()
         self.last_journal_msg_time = time.monotonic()
         self.searches = list(self._build_searches(searches))
@@ -207,21 +218,12 @@ class JournalReader(Tagged):
             self.journald_reader.close()
 
     def initialize_senders(self):
-        if self._senders_initialized:
-            return
-
-        senders = {
-            "elasticsearch": ElasticsearchSender,
-            "kafka": KafkaSender,
-            "logplex": LogplexSender,
-            "file": FileSender,
-            "rsyslog": RsyslogSender,
-            "aws_cloudwatch": AWSCloudWatchSender,
-            "google_cloud_logging": GoogleCloudLoggingSender,
-        }
-        for sender_name, sender_config in self.config.get("senders", {}).items():
+        configured_senders = self.config.get("senders", {})
+        for sender_name, sender_config in configured_senders.items():
+            if sender_name in self._initialized_senders:
+                continue
             try:
-                sender_class = senders[sender_config["output_type"]]
+                sender_class = self.sender_classes[sender_config["output_type"]]
             except KeyError as ex:
                 raise Exception("Unknown sender type {!r}".format(sender_config["output_type"])) from ex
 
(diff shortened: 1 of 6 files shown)


journalpump/senders/file.py
```diff
@@ -5,7 +5,7 @@ class FileSender(LogSender):
     def __init__(self, *, config, **kwargs):
         super().__init__(config=config, max_send_interval=config.get("max_send_interval", 0.3), **kwargs)
         self.mark_disconnected()
-        self.output = open(config["file_output"], "ab")
+        self.output = open(config["file_output"], "ab")  # pylint: disable=consider-using-with
```
deztructor, line 8: I think pylint warnings should be handled in a different commit. Also, do these pylint warnings mean there is technical debt in the implementation?

jhonkola: Pylint 2.8, released at 2021-04-24, added a new refactoring message "consider-using-with" (http://pylint.pycqa.org/en/latest/whatsnew/2.8.html). The checker generates these messages
for roughly all calls that are often used with context managers. These cases seemed to me such
that it does not make sense to start changing them, so I added the directives here to get the code
to pass the linter.

jhonkola: Yes, in `FileSender` the `request_stop()` should close the opened file as it is not done elsewhere. Good catch.

jhonkola: This would be a good idea. I think that the log file handling could be put inside a context manager by overriding `run()` in `FileSender`.

journalpump/journalpump.py
```diff
@@ -207,21 +218,12 @@ def close(self):
...
-            "google_cloud_logging": GoogleCloudLoggingSender,
-        }
-        for sender_name, sender_config in self.config.get("senders", {}).items():
+        configured_senders = self.config.get("senders", {})
+        for sender_name, sender_config in configured_senders.items():
+            if sender_name in self._initialized_senders:
+                continue
             try:
-                sender_class = senders[sender_config["output_type"]]
+                sender_class = JournalReader.sender_classes[sender_config["output_type"]]
```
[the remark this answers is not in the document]
jhonkola: Fixed

journalpump/journalpump.py
```diff
@@ -233,21 +235,29 @@ def initialize_senders(self):
...
+                    name=sender_name,
+                    reader=self,
+                    stats=self.stats,
+                    tags=self.make_tags(),
+                )
+            except SenderError:
+                # If sender init fails, log exception, don't start() the sender
+                # and don't add it to self.senders dict. A metric about senders that
+                # failed to start is sent at the end of this method
+                self.log.exception("Sender %r failed to initialize", sender_name)
```
tsharju, line 253: I think it would be good to retry at least few times before deciding the sender has failed. This might be a bit sender specific but at least the AWS sender uses `LogSender._backoff` to throttle retries and this is used if there are errors while sending the log messages upstream. I think the sending of the log messages is retried indefinitely (sleeping max 30 seconds between retries), but here it could maybe retry a few times and then decide the sender has failed.

tsharju (changes requested):
Looks good overall. One comment about retrying the initialization of the sender few times before giving up on it.

journalpump/journalpump.py
tsharju: Ah, sorry. It is raising `SenderError` only if authentication fails. So maybe no reason to retry that. Maybe rename the `SenderError` to something less generic like `SenderInitializationError`?

tsharju: If it gets pass the `LogSender.start` method the retrying will actually happen when sending messages.

jhonkola: The authentication failure happens in `AWSCloudWatchSender(LogSender).__init__()` so `LogSender.run()` has not been called yet. Adding a few retries with `LogSender._backoff()` would still be easy to do in `AWSCloudWatchSender._init_logs()`. AWS does return transient 503 errors occasionally, so retrying would handle this case nicely as well.

I'll rename `SenderError` to `SenderInitializationError` to avoid confusion.

journalpump/senders/aws_cloudwatch.py
```diff
@@ -27,17 +29,46 @@ def _init_logs(self):
...
+                        raise SenderInitializationError from err
+                    continue
+                # If we get some other aws exception, log it and raise if too many attempts.
+                # This is not handled in any special way
+                self.log.exception("AWS ClientError")
+                self._backoff()
+                if no_more_attempts:
+                    raise
+            else:
+                success = True
```
deztructor, line 71: Maybe you can use `break` instead of `success` variable.

deztructor (dismissed):
Some comments.
Also it could be beneficial for the history to split the change to >1 commits, e.g. I believe retrying could be a separate change.

journalpump/senders/aws_cloudwatch.py
```diff
@@ -27,17 +29,46 @@ def _init_logs(self):
...
+            no_more_attempts = attempt >= MAX_INIT_TRIES
+            try:
+                # Create the log group and stream if they don't exist yet
+                try:
+                    self._logs.create_log_group(logGroupName=self.log_group)
+                except botocore.exceptions.ClientError as err:
+                    if err.response["Error"]["Code"] != "ResourceAlreadyExistsException":
+                        raise
+                try:
+                    self._logs.create_log_stream(logGroupName=self.log_group, logStreamName=self.log_stream)
```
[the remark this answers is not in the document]
jhonkola: Combining these will not work in a scenario where log group exists, and log stream does not. `create_log_group()` will raise `ClientError` always when called, jumping to `except`, and then `create_log_stream()` will not get called.

journalpump/senders/aws_cloudwatch.py
```diff
@@ -27,17 +29,46 @@ def _init_logs(self):
...
-        except botocore.exceptions.ClientError as err:
-            if err.response["Error"]["Code"] != "ResourceAlreadyExistsException":
-                raise
+
+        # Catch access denied exception(e.g. due to erroneous credentials)
+        attempt = 0
+        success = False
+        while attempt < MAX_INIT_TRIES and not success:
+            attempt += 1
+            no_more_attempts = attempt >= MAX_INIT_TRIES
```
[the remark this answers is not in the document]
jhonkola: Good idea, thanks.

tsharju (approved):

PR #91, opened 2021-08-11
websocket: Add a websocket sender

aihessu:
Add support for sending messages to a websocket server.


journalpump/senders/websocket.py
```diff
@@ -0,0 +1,286 @@
...
+from threading import Thread
+from urllib.parse import urlparse
+
+import asyncio
+import logging
+import random
+import socket
+import ssl
+import time
+import websockets
```
rikonen, line 15: This isn't built-in package but I don't see any dependency declared anywhere.

journalpump/senders/websocket.py
```diff
@@ -0,0 +1,286 @@
...
+            base=10, factor=1.8, maximum=60, jitter=True
+        )
+        # prevent websockets from logging message contents when we're otherwise in DEBUG mode
+        logging.getLogger("websockets").setLevel(logging.INFO)
+
+        self.socks5_proxy = None
+        if self.socks5_proxy_url:
+            self.socks5_proxy = Proxy.from_url(self.socks5_proxy_url, loop=self.websocket_loop)
+
+    async def consumer_handler(self, websocket):
```
rikonen, line 80: Journalpump at the moment doesn't have any async code. Did you check if there are any good quality sync websocket clients that could be used instead? Since the same model than what you have here is successfully used elsewhere too I don't think it's a problem as such but just seems like synchronous client might have been more suitable in journalpump context. Now the async handling just adds some inefficiences when switching contexts and makes the code more complex.

journalpump/senders/websocket.py
```diff
@@ -0,0 +1,286 @@
...
+        # Dummy consumer to read and ignore messages from the websocket
+        while self.running:
+            _ = await websocket.recv()
+
+    async def producer_handler(self, websocket):
+        while self.running:
+            messages = await self.queue.async_q.get()
+
+            for message in messages:
```
rikonen, line 89: Going forward probably makes sense to add batch support so that these can all be sent in one request. Possibly also some configurable minimum send interval so that there won't be 1000 requests per second if there are 1000 evenly timed log messages per second.

journalpump/senders/websocket.py
```diff
@@ -0,0 +1,286 @@
+from .base import LogSender
+from aiohttp_socks import ProxyConnectionError, ProxyError, ProxyTimeoutError
+from aiohttp_socks.utils import Proxy
+from concurrent.futures import TimeoutError as ConnectionTimeoutError
+from janus import Queue
```
rikonen, line 5: This isn't built-in library but there's no dependency to this. Also I would not add this dependency at all. Looking at the implementation it doesn't seem to do anything magical / especially efficient. You can just make `send()` do `asyncio.run_coroutine_threadsafe(self.async_send(messages), self.asyncio_loop).result()` and work with regular queue along with asyncio locks.

journalpump/senders/websocket.py
```diff
@@ -0,0 +1,286 @@
...
+        await self.stop_event.wait()
+
+    async def websocket_connect(self, *, timeout=40):
+        ssl_context = None
+        if self.ssl_enabled:
+            ssl_context = ssl.SSLContext(ssl.PROTOCOL_TLS_CLIENT)
+            ssl_context.load_cert_chain(self.ssl_cert, keyfile=self.ssl_key)
+            ssl_context.load_verify_locations(self.ssl_ca)
+
+        headers = {"User-Agent": "journalpump"}
```
rikonen, line 128: Having version here might be nice

journalpump/senders/websocket.py
```diff
@@ -0,0 +1,286 @@
...
+                )
+                runner.start()
+            except Exception as ex:  # pylint:disable=broad-except
+                self.mark_disconnected(ex)
+                self.log.exception("Retriable error during Websocket initialization: %s: %s", ex.__class__.__name__, ex)
+                self._backoff()
+            else:
+                self.log.info("Initialized Websocket client, address: %r", self.config["websocket_uri"])
+                self.runner = runner
+                self.mark_connected()
```
rikonen, line 262: It would be nicer to report the actual connected status. Now this just reports whether the thread has been created which in practice always succeeds and is always running yet connection might well not exist.

rikonen (changes requested):

journalpump/senders/websocket.py
aihessu: Yes indeed, the idea is to batch individual log messages and do end-to-end compression from here to kafkapump in a way that nodecom does not need to decompress and open up the batches. But that shall happen in the next iteration, at this point I'll just make the basic thing work.

journalpump/senders/websocket.py
aihessu: Yes, absolutely. Placeholder to get things running.

journalpump/senders/websocket.py
aihessu: There's one which meets the basic requirements (active project, SOCKS5 support) at https://github.com/websocket-client/websocket-client but it does run another thread as well to manage the connection, so the context switching will still be around. Also has a bug from 2019 where it gets stuck and does not recover after a network connectivity failure (https://github.com/websocket-client/websocket-client/issues/592), although there might be a fix found for it now. An old version is packaged in Fedora, we'd need to package it ourself (and likely a couple of dependencies). Would feel safer to go with the already-familiar client at this point and reconsider later if necessary.

journalpump/senders/websocket.py
aihessu: Yes indeed

journalpump/senders/websocket.py
aihessu: I ended up removing the, as journalpump does not seem to require one here at this time.

rikonen (changes requested):

journalpump/senders/websocket.py
```diff
@@ -0,0 +1,282 @@
...
+        self.setDaemon(True)
+        self.log = logging.getLogger(self.__class__.__name__)
+        self.websocket_uri = websocket_uri
+        self.socks5_proxy_url = socks5_proxy_url
+        self.ssl_enabled = ssl_enabled
+        self.ssl_ca = ssl_ca
+        self.ssl_key = ssl_key
+        self.ssl_cert = ssl_cert
+        self.websocket_loop = asyncio.new_event_loop()
+        asyncio.set_event_loop(self.websocket_loop)
```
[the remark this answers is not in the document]
rikonen: Interesting. I wonder what the recommended way of creating events is in Python 3.10 then. This seems like totally valid use-case. Setting the loop here at least seems clearly incorrect so creating the events from the other thread might be next best option.

journalpump/senders/websocket.py
```diff
@@ -0,0 +1,282 @@
...
+import socket
+import ssl
+import time
+import websockets
+
+
+class WebsocketRunner(Thread):
+    def __init__(self, *, websocket_uri, socks5_proxy_url, ssl_enabled, ssl_ca, ssl_key, ssl_cert):
+        super().__init__()
+        self.setDaemon(True)
```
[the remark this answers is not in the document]
aihessu: It is currently stopped at shutdown, this is mostly a precaution in case something goes wrong in the shutdown. I would leave it in.

PR #92, opened 2021-08-16
Also add reference to Ubuntu

derrix060:
Since Ubuntu is one of the most popular distro, there are people with different knowledge that uses it.  Some people won't realize that the instructions for Debian are also valid for Ubuntu


jankatins:
Can you confirm that you tested the dpkg -i on ubuntu? Just making sure that all dependencies are actually available in ubuntu...

derrix060:
I didn't test

PR #93, opened 2021-08-24
journal: add log namespace support [BF-68]

alexole:


README.rst
```diff
@@ -260,6 +260,10 @@ which the journalpump feeds the configured logsender.
 Path to the directory containing journal files if you want to override the
 default one.
 
+``journal_namespace`` (default ``null``)
+
+Journal namespace to read logs from.
```
jankatins, line 265: Can you say what will be read in case this ends up null/None? `If null/unset, will read ....`

alexole: Default behavior of python-systemd, we are just passing over the parameters to Reader and I hope this is not going to change. Also in the context it should be pretty clear that if it's not provided, should be the same behaviour as for journalctl and other tools - read all, no filtering.

jankatins: I had two hypotheses: either "all" or "the default namespace", was the first time I heard about namespaces :-). 

I still think it would be good to write it here it as we might not be the only user of journalpump and it's bad if you have to read the source code to verify what it means.

jankatins (dismissed):

jankatins (approved):

PR #95, opened 2021-09-07
journal: create journal reader only when journal files are available [BF-58]

alexole:
Wait for the journal files (`/var/log/journal/<journal>.<namespace>/journal` or `/run/..` ) to appear before creating the readers, otherwise in case of file is not available the readers will start, but won't be able to read the logs. It's not a problem usually for default journal namespace, because it's usually created way before the `journalpump` starts, however if you use e.g. some namespace which does not yet exist in journal (will be created later), reader won't be able to process data.

### DIFF
--- a/test/test_journalpump.py
+++ b/test/test_journalpump.py
@@ -2,7 +2,7 @@ from .data import GCP_PRIVATE_KEY
 from botocore.stub import Stubber
 from collections import OrderedDict
 from datetime import datetime
-from journalpump.journalpump import FieldFilter, JournalObject, JournalObjectHandler, JournalPump, JournalReader
+from journalpump.journalpump import FieldFilter, JournalObject, JournalObjectHandler, JournalPump, JournalReader, PumpReader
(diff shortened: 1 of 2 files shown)


test/test_journalpump.py
```diff
@@ -49,7 +49,8 @@ def test_journalpump_init(tmpdir):  # pylint: disable=too-many-statements
     assert len(a.readers) == 1
     for rn, r in a.readers.items():
         assert rn == "foo"
-        r.create_journald_reader_if_missing()
+        with mock.patch.object(PumpReader, "has_persistent_files", return_value=True):
```
alexole, line 52: I'd prefer patching it in decorator for the test method, but for some reason mock library leaves some temporary files `<MagicMock name=.....>' without properly cleaning up the mocking. Maybe due to multithreading, checked with joins - it does not help. I guess I can leave it like that.

jankatins (approved):
LGTM

PR #100, opened 2021-12-09
Recreate reader if journal files is deleted

m-novikov:
To allow this changes we always keep journal reader registered with
poller. For every even received we keep a lookup table of said events,
that should be processed at later stage.

There are 2 relevant types of events:
* journal.APPEND - when lines are added to the end of the file
* journal.INVALIDATE - log file rotated or removed

For APPEND event we need to read all the data in the log until it's
exhausted

For INVALIDATE event data should be also read until the end and the
reader should be recreated to release fds of deleted files.

This fixes resource leak which happened if log file was first
renamed and then removed, which usually is the case when rotation
happens.


journalpump/journalpump.py
```diff
@@ -727,28 +800,88 @@ def _close_stale_readers(self):
...
+
+                # Read messages up to set reader limit
+                exhausted = False
+                if event in [journal.APPEND, journal.INVALIDATE]:
+                    exhausted, line_count = self.read_messages(reader, hits, chunk_size=5000)
+                    lines += line_count
+
+                if event == journal.INVALIDATE:
+                    self.unregister_from_poll(reader)
+                    reader.invalidate()
```
fingon, line 874: If this isn't exhausted, doesn't this mean it doesn't read all? Should this be done only if exhausted?

fingon (dismissed):
Using namedtuple (or frozen dataclass) would be better for the return values too, but didn't comment about them explicitly. However, this one seems like potential bug or I misunderstand the logic.

fingon (approved):

PR #109, opened 2022-02-03
journalpump: add support for unit_log_levels

alxric:
This will allow specifying log level limits per unit basis in case you want to filter what log level severities journalpump picks up

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1,5 +1,5 @@
 journalpump |BuildStatus|_
-========================
+==========================
 
 .. |BuildStatus| image:: https://github.com/aiven/journalpump/actions/workflows/build.yml/badge.svg?branch=master
 .. _BuildStatus: https://github.com/aiven/journalpump/actions
@@ -106,6 +106,9 @@ Example::
       "field_filters": {
          ...
       },
+      "unit_log_levels": {
+         ...
+      },
       "json_state_file_path": "/var/lib/journalpump/journalpump_state.json",
       "readers": {
          ...
@@ -170,6 +173,36 @@ excluded (``blacklist``).
 The actual fields to include or exclude. Field name matching is case
 insensitive and underscores in the beginning of the fields are trimmed.
 
+Unit log levels configuration
+=============================
+
+Unit log levels can be used to specify which log levels you want to set on a per unit basis. Matching supports glob
+patterns. For example, to only process messsages for a systemd-unit called ``test-unit`` with severity ``WARNING`` or higher,
+your config could look like this::
+
+  {
+      "unit_log_levels": {
+          "log_level_name": [
+              {
+                  "service_glob": "test-unit*",
+                  "log_level": "WARNING"
+              },
+              {
+                  "service_glob": "*-unit",
+                  "log_level": "INFO"
+              }
+          ]
+      }
+  }
+
+Note that if your unit would match multiple patterns (like "test-unit" would in the example above), the first match will
+get used, i.e "WARNING" in this case.
+
+``log_level_name``
+
+Name of the log level configuration. This can be configured per sender and depending
+on the use case the settings for different senders may vary.
+
 Reader configuration
 ====================
 Reader configuration structure::
@@ -205,11 +238,20 @@ Example configuration for a single reader::
               "type": "blacklist"
           }
       },
+      "unit_log_levels": {
+          "drop_everything_below_warning": [
+              {
+                  "service_glob": "*",
+                  "log_level": "WARNING"
+              }
+          ]
+      },
       "journal_path": "/var/lib/machines/container1/var/log/journal/b09ffd62229f4bd0829e883c6bb12c4e",
       "senders": {
           "k1": {
               "output_type": "kafka",
               "field_filter": "drop_process_id",
+              "unit_log_level": "drop_everything_below_warning",
               "ca": "/etc/journalpump/ca-bundle.crt",
               "certfile": "/etc/journalpump/node.crt",
               "kafka_address": "kafka.somewhere.com:12345",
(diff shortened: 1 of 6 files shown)


ryanisroot (commented):
This looks good to me, tested with it a little locally

test/test_journalpump.py
```diff
@@ -358,25 +359,66 @@ def test_blacklist(self):
...
+        log_level = "WARNING"
+        ll = UnitLogLevel("test", {"unit-a": log_level})
+        data = {"Foo": "bar", "PRIORITY": LOG_SEVERITY_MAPPING[log_level] + 1, "SYSTEMD_UNIT": "unit-a"}
+        assert ll.filter_by_level(data) == {}
+
+    def test_lower_level(self):
+        log_level = "WARNING"
+        ll = UnitLogLevel("test", {"unit-a": log_level})
+        data = {"Foo": "bar", "PRIORITY": LOG_SEVERITY_MAPPING[log_level] - 1, "SYSTEMD_UNIT": "unit-a"}
+        assert ll.filter_by_level(data) == data
```
mble, line 385: Could we have a couple of test cases to match when `SYSTEMD_UNIT` and/or `PRIORITY` are not present in the entry? We should be allowing these though regardless of the filter.

alxric: We can indeed! Added

mble (approved):
LGTM.

mble:
I'm not merging in case anyone else would like to take a look.

alxric:
Made some changes to match unit names via glob pattern instead of exact name as per suggestion by @fingon

journalpump/journalpump.py
```diff
@@ -542,6 +550,35 @@ def filter_fields(self, data):
...
+        priority = entry.get("PRIORITY")
+        if ((not priority) or (not unit) or ((self.known_units.get(unit, False) or self._unit_match_level_glob(unit))
+                                             and priority <= LOG_SEVERITY_MAPPING[self.levels[unit]])):
+            self.known_units[unit] = True
+            return data
+        self.known_units[unit] = False
+        return {}
+
+    def _unit_match_level_glob(self, unit: str) -> bool:
+        for unit_pattern, level in self.levels.items():
```
fingon, line 575: This order is semi-arbitrary (Python specifies dicts are sorted but you do not document it that way in the readme, and I think relying on it might be dangerous if there's e.g. json serialization/deserialization in the middle). I would recommend switching to a list to be more explicit, and it should be documented that it is first match that is used.

journalpump/journalpump.py
```diff
@@ -542,6 +550,35 @@ def filter_fields(self, data):
...
+            if fnmatch.filter([unit], unit_pattern):
```
fingon, line 576:
```suggestion
            if fnmatch.fnmatch(unit, unit_pattern):
```

fingon (dismissed):
Mostly LGTM. Comment or two about configuration format.

fingon (changes requested):
I guess I should have perhaps written it out but I was after just list of dicts with service name glob key, and level to use for it key. We can easily extend that later if necessary by adding more key+values to those dicts, and the current service unit match = log level dict is not similarly extensible. 

( eg someone might want to comment them for some reason )

README.rst
```diff
@@ -170,6 +173,31 @@ excluded (``blacklist``).
...
+Unit log levels configuration
+=============================
+
+Unit log levels can be used to specify which log levels you want to set on a per unit basis. Matching supports glob
+patterns. For example, to only process messsages for a systemd-unit called ``test-unit`` with severity ``WARNING`` or higher,
+your config could look like this::
+
+  {
+      "unit_log_levels": {
+          "log_level_name": [
```
fingon, line 185:
```suggestion
      "unit_log_levels": [
```

(Toplevel README already said it was a list)

journalpump/journalpump.py
```diff
@@ -542,6 +550,42 @@ def filter_fields(self, data):
...
+        if self.known_units[unit] is not None and priority <= LOG_SEVERITY_MAPPING[self.known_units[unit]]:
+            return data
+
+        return {}
+
+    def _unit_match_level_glob(self, unit: str):
+        self.known_units[unit] = None
+        for level in self.levels:
+            if fnmatch.fnmatch(unit, level.get("service_glob", "")):
+                self.known_units[unit] = level.get("log_level")
```
fingon, line 584:
```suggestion
                self.known_units[unit] = level["log_level"]
```

I think defaulting to None leads to undesirable results.

journalpump/journalpump.py
```diff
@@ -650,6 +709,10 @@ def configure_field_filters(self):
         filters = self.config.get("field_filters", {})
         self.field_filters = {name: FieldFilter(name, config) for name, config in filters.items()}
 
+    def configure_unit_log_levels(self):
+        unit_log_levels = self.config.get("unit_log_levels", {})
```
fingon, line 713:
```suggestion
        unit_log_levels = self.config.get("unit_log_levels", [])
```

fingon (dismissed):
LGTM now - 3 minor tweaks still.

README.rst
alxric: I will fix top level README instead as I think this is how it should be. I made it like this to match the behaviour of `field_filters` where you name the `field_filter` so that later in the `reader` section you speficy the name of the `field_filter` you want to use, so that you can use different field filters for different readers.

journalpump/journalpump.py
alxric: see comment above. I think this is how we want it to be

fingon (approved):

PR #112, opened 2022-02-16
Add a secret filtering functionality to journalpump.

ryanisroot:
This change adds a secret filtering functionality. Config is at the reader level to allow one reader to redact secrets for multiple senders, but enable different behaviour between readers.
Provide a regex with groups, a redaction message, and define which groups are to be redacted in the case of a match. Multiple patterns can be added.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -268,6 +268,12 @@ Example configuration for a single reader::
               "name": "journal.oom_killer"
           }
       ],
+      "secret_filter_metrics": true,
+      "secret_filters": [
+        {
+          "pattern": "SENSITIVE",
+          "replacement": "[REDACTED]"
+        }],
       "tags": {
           "type": "container"
       }
@@ -319,6 +325,49 @@ and ``"SYSTEM"`` opens journal files of system services and the kernel, ``"CURRE
 current user; and ``"OS_ROOT"`` is used to open the journal from directories relative to the specified
 directory path or file descriptor. Multiple flags can be OR'ed together using a list:
 ``["LOCAL_ONLY", "CURRENT_USER"]``.
+
+``secret_filters`` (default ``[]``)
+
+Secret filters can be used to redact sensitive data which matches known patterns in logs before forwarding the message along
+to it's final destination. To use: add a number of filters following the pattern below to the reader config. The ``pattern`` is a standard
+python regex, and the matching substring will be subbed with ``replacement``. Patterns are compiled at runtime.
+
+Simple pattern example:
+
+This simple pattern should be used for most cases. It will replace SECRET with [REDACTED] but will leave the rest of the message intact.
+
+"secret_filters": [
+  {
+    "pattern": "SECRET",
+    "replacement": "[REDACTED]"
+  }
+]
+
+Complex pattern example:
+
+For more complex requirements, a python regex with capture groups can be provided, and the contents of the message restructured using backrefs.
+This example will only replace SENSITIVE with [REDACTED] as long as foo and bar are also part of the pattern.
+
+"secret_filters": [
+  {
+    "pattern": "(bar)(SENSITIVE)(foo)",
+    "replacement": "\\1[REDACTED]\\3",
+  }
+]
+
+Using backrefs, the message can also be restructured into a new format.
+"secret_filters": [
+  {
+    "pattern": "(bar)(SENSITIVE)(foo)",
+    "replacement": "\\1\\3 pattern was [REDACTED]",
+  }
+]
+
+
+``secret_filter_metrics`` ( default: ``false``)
+Change this setting to true to emit metrics to the metrics host whenever a secret pattern is matched.
+This matching happens before other filtering to help catch secrets being leaked to disk.
+
 
 
 Sender Configuration
--- a/journalpump/journalpump.py
+++ b/journalpump/journalpump.py
@@ -173,6 +173,10 @@ class JournalReader(Tagged):
         self.last_stats_send_time = time.monotonic()
         self.last_journal_msg_time = time.monotonic()
         self.searches = list(self._build_searches(searches))
+        self.secret_filter_matches = 0
+        self.secret_filters = self._validate_and_build_secret_filters(config)
+        self.secret_filter_metrics = self._configure_secret_filter_metrics(config)
+        self.secret_filter_metric_last_send = time.monotonic()
         self._is_ready = True
 
     def invalidate(self) -> None:
@@ -330,6 +334,15 @@ class JournalReader(Tagged):
             "total_bytes": self.read_bytes,
         }
 
+    def report_secret_filter_metrics(self):
+        now = time.monotonic()
+        if (now - self.last_stats_send_time) < 10.0:
+            return
+        tags = self.make_tags()
+        self.stats.increase("journal.secret_filter_match", self.secret_filter_matches, tags=tags)
+        self.secret_filter_matches = 0
+        self.secret_filter_metric_last_send = now
+
     def inc_line_stats(self, *, journal_lines, journal_bytes):
         self.read_bytes += journal_bytes
         self.read_lines += journal_lines
@@ -343,6 +356,7 @@ class JournalReader(Tagged):
         self.stats.gauge("journal.last_read_ago", value=now - self.last_journal_msg_time, tags=tags)
         self.stats.gauge("journal.read_lines", value=self.read_lines, tags=tags)
         self.stats.gauge("journal.read_bytes", value=self.read_bytes, tags=tags)
+
         self.last_stats_send_time = now
         self._sent_lines_diff += self.read_lines - self._last_sent_read_lines
         self._sent_bytes_diff += self.read_bytes - self._last_sent_read_bytes
@@ -495,6 +509,36 @@ class JournalReader(Tagged):
 
             yield output
 
+    def _configure_secret_filter_metrics(self, config):
+        secret_filter_metrics = config.get("secret_filter_metrics")
+        if secret_filter_metrics is True:
+            return True
+        return False
+
+    def _validate_and_build_secret_filters(self, config):
+        secret_filters = config.get("secret_filters")
+        if secret_filters is None:
+            return []
+
+        if not isinstance(secret_filters, list):
+            raise ValueError("invalid secret_filters configuration - must be a list")
+
+        for secret_filter in secret_filters:
+            if secret_filter.get("replacement") is None:
+                raise ValueError("invalid secret_filters configuration - missing field 'replacement'")
+
+            if secret_filter.get("pattern") is None:
+                raise ValueError("invalid secret_filters configuration - missing field 'pattern'")
+
+        # Compile / validate regex
+        for idx, sfilter in enumerate(secret_filters):
+            try:
+                secret_filters[idx]["compiled_pattern"] = re.compile(sfilter["pattern"])
+            except re.error as e:
+                raise ValueError("invalid secret_filters configuration - invalid regex") from e
+
+        return secret_filters
+
     def perform_searches(self, jobject):
         entry = jobject.entry
         results = {}
@@ -608,6 +652,10 @@ class JournalObjectHandler:
         if not self.pump.check_match(new_entry):
             return SingleMessageReadResult(has_more=True, bytes_read=None)
 
+        # Redact secrets before other filtering for effective leak detection metrics
+        if self.reader.secret_filters:
+            new_entry = self._apply_secret_filters(new_entry)
+
         for sender in self.reader.senders.values():
             json_entry = self._get_or_generate_json(
                 sender.field_filter, sender.unit_log_levels, sender.extra_field_values, new_entry
@@ -674,6 +722,18 @@ class JournalObjectHandler:
             "partial_data": json_entry[:1024],
         }
         return json.dumps(entry, default=default_json_serialization).encode("utf8")
+
+    def _apply_secret_filters(self, data):
+        msg = data.get("MESSAGE")
+        original_msg = msg
+        if msg:
+            for secret_filter in self.reader.secret_filters:
+                msg = secret_filter["compiled_pattern"].sub(secret_filter["replacement"], msg)
+        if msg != original_msg:
+            if self.reader.secret_filter_metrics:
+                self.reader.secret_filter_matches += 1
+            data["MESSAGE"] = msg
+        return data
 
 
 class JournalPump(ServiceDaemon, Tagged):
@@ -910,7 +970,7 @@ class JournalPump(ServiceDaemon, Tagged):
             self.reader_by_fd[fd] = self._STALE_FD
             self.log.info("Unregistered reader %r with fd %r", self.name, fd)
 
-    def run(self):
+    def run(self):  # pylint: disable=too-many-statements
         last_stats_time = 0
         poll_timeout = 0
         buffered_events = {}
@@ -976,6 +1036,9 @@ class JournalPump(ServiceDaemon, Tagged):
                 if self.register_for_poll(reader):
                     # Check if there is something to read in newly created reader
                     buffered_events[reader] = journal.APPEND
+
+                if reader.secret_filter_metrics:
+                    reader.report_secret_filter_metrics()
 
             if hits and time.monotonic() - last_stats_time > 60.0:
                 self.log.info("search hits stats: %s", hits)
--- a/test/test_journalpump.py
+++ b/test/test_journalpump.py
@@ -18,6 +18,7 @@ from unittest import mock, TestCase
 import boto3
 import json
 import pytest
+import re
 import responses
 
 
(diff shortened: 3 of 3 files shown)


journalpump/journalpump.py
```diff
@@ -495,6 +496,14 @@ def value_func(tags, f=f, args=args):  # pylint: disable=undefined-variable
 
             yield output
 
+    def _build_secret_filters(self, config):
+        secret_filters = config.get("secret_filters")
+        if secret_filters is not None:
+            for idx, _ in enumerate(secret_filters):
+                secret_filters[idx]["compiled_pattern"] = re.compile(secret_filters[idx]["pattern"])
```
mble, line 503: This is optimistic dict access, with an invalid config raising an exception. Need to either validate the `secret_filters` entry has a valid structure, use safe access, or a `try`/`except KeyError` block to handle the possibility.

journalpump/journalpump.py
```diff
@@ -675,6 +687,24 @@ def _truncate_long_message(self, json_entry):
...
+    def _apply_secret_filters(self, data):
+        changed = False
+        for idx, _ in enumerate(self.reader.secret_filters):
+            msg = data["MESSAGE"]
+            m = self.reader.secret_filters[idx]["compiled_pattern"].match(msg)
+            if m:
+                msg = ""
+                for idy, g in enumerate(m.groups()):
+                    if idy in self.reader.secret_filters[idx]["redact_groups"]:
+                        msg += self.reader.secret_filters[idx]["redaction_message"]
```
mble, line 699: Same here with dict access.

journalpump/journalpump.py
```diff
@@ -495,6 +496,14 @@ def value_func(tags, f=f, args=args):  # pylint: disable=undefined-variable
 
             yield output
 
+    def _build_secret_filters(self, config):
+        secret_filters = config.get("secret_filters")
+        if secret_filters is not None:
+            for idx, _ in enumerate(secret_filters):
```
mble, line 502: Prefer `for secret_filter in secret_filters`.

journalpump/journalpump.py
```diff
@@ -495,6 +496,14 @@ def value_func(tags, f=f, args=args):  # pylint: disable=undefined-variable
 
             yield output
 
+    def _build_secret_filters(self, config):
+        secret_filters = config.get("secret_filters")
```
mble, line 500: Could use `config.get("secret_filters", [])`. Can then check the type of `secret_filters` to ensure it is a list before iteration.

journalpump/journalpump.py
```diff
@@ -675,6 +687,24 @@ def _truncate_long_message(self, json_entry):
         }
         return json.dumps(entry, default=default_json_serialization).encode("utf8")
 
+    def _apply_secret_filters(self, data):
+        changed = False
+        for idx, _ in enumerate(self.reader.secret_filters):
```
mble, line 692: Same here, preference for `for secret_filter in self.reader.secret_filters`, as `idx` is only being used to access that list and not a correlated list.

journalpump/journalpump.py
```diff
@@ -675,6 +687,24 @@ def _truncate_long_message(self, json_entry):
...
+            msg = data["MESSAGE"]
+            m = self.reader.secret_filters[idx]["compiled_pattern"].match(msg)
```
mble, line 694: Mild preference for the var to be called `match` over `m`.

journalpump/journalpump.py
mble: Need to also protect against an invalid pattern being provided when compiling.

mble (commented):
Left some comments, mostly nits about dict access and iteration.

As the config gets more complex, we want to ensure that we have safety mechanisms to prevent invalid config from crashing `journalpump`, falling back to "safe defaults" where it is appropriate to do so.

journalpump/journalpump.py
ryanisroot: Agree everywhere else, but in this case I am modifying the original object. 
So, I could use this pattern + build a new object, or stick with the existing pattern. Any preference?

journalpump/journalpump.py
ryanisroot: What would the ideal behavior be in the case of an invalid regex? 

I think I should add a try catch, but mostly for a neater looking error message. An invalid regex in the config should probably still exit journalpump. The alternative would be dropping the config and running in a degraded state (without filtering) but I think that's a bad pattern as it lets the issue go unnoticed.

mble: Hmm. I think mostly with a resiliency cap on - I'd rather run in a degraded state versus not running at all, but can see the benefits of simply crashing. I'll +1 whatever you decide on this, but at least giving a nice error message would be good :)

journalpump/journalpump.py
ryanisroot: In your example, i'd be modifying a copy of the list item, not the item in the original list. 
Here's a contrived example to show what i mean:
```
>>> list_of_items = [1, 2, 3, 4, 5]
>>> 
>>> # operating on a copy of the list item
>>> for item in list_of_items:
...     item = item + 1
... 
>>> print(list_of_items)
[1, 2, 3, 4, 5]
>>> 
>>> # operating on the original list object
>>> for idx, _ in enumerate(list_of_items):
...     list_of_items[idx] = list_of_items[idx] + 1
... 
>>> print(list_of_items)
[2, 3, 4, 5, 6]
```

journalpump/journalpump.py
ryanisroot: I'm going to wear my security cap here and lean towards saying it would be better to send nothing at all, than to think you are protected by filtering which was disabled. In the case of something like an API service, I'd probably rather have partly working than completely broken.

journalpump/journalpump.py
```diff
@@ -647,6 +656,9 @@ def _get_or_generate_json(self, field_filter, unit_log_levels, extra_field_value
         if field_filter:
             data = field_filter.filter_fields(data)
 
+        # Redact secrets before truncating
+        data = self._apply_secret_filters(data)
```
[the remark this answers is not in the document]
ryanisroot: Great suggestion. 

I was originally thinking to do the regex matching at late as possible ( but after truncating with the idea that maybe redacting a secret might help retain a valid payload structure in some cases ) to minimize the compute impact, but now that you mention it, emitting metrics on any secret patterns leaked at all seems more valuable than some cpu cycles.

journalpump/journalpump.py
ryanisroot: resolved with a config validation step + removed idx in cases where it's not really needed ( read only loops )

journalpump/journalpump.py
```diff
@@ -675,6 +718,32 @@ def _truncate_long_message(self, json_entry):
...
+
+        for secret_filter in self.reader.secret_filters:
+            msg = data.get("MESSAGE")
+            if msg:
+                match = secret_filter["compiled_pattern"].match(msg)
+                if match:
+                    msg = ""
+                    changed = True
+
+                    self.pump.stats.increase(
```
ryanisroot, line 732: converted to draft because i forgot to put this metric emission behind a config option as requested.

README.rst
```diff
@@ -320,6 +328,32 @@ current user; and ``"OS_ROOT"`` is used to open the journal from directories rel
...
+to it's final destination. To use: add a number of filters following the pattern below to the reader config. The ``pattern`` is a standard
+python regex with groups. ``redact_groups`` defines which groups contents will be replaced, with the ``redaction_message``. Patterns are
+compiled at runtime.
+
+
+"secret_filters": [
+  {
+    "pattern": "(.*?\\s?)(SECRET)(\\s?.*)",
+    "redact_groups": [1],
+    "redaction_message": "[REDACTED]"
```
rikonen, line 343: Not sure if this was discussed somewhere but was there some particular reason for going with this sort of syntax? For simple cases this seems pretty complex and also error prone as your lines will just be truncated if you forgot the (.*) after the string you're actually looking for. Pattern to look for and allowing referencing match groups in redaction message for more advanced use cases might be easier to use. I.e. the above would be expressed with

```
{
    "pattern": "SECRET",
    "replacement": "[REDACTED]"
}
```

Then you could also do something like

```
{
    "pattern": "(some prefix) SECRET (followed by)",
    "replacement": "\1 [REDACTED] \2"
}
```

To do "This is some prefix SECRET followed by something" => "This is some prefix [REDACTED] followed by something". Also allows more complex transformations as needed and implementation is just `re.sub(pattern, replacement, log_line)`

journalpump/journalpump.py
```diff
@@ -675,6 +725,33 @@ def _truncate_long_message(self, json_entry):
...
+        for secret_filter in self.reader.secret_filters:
+            msg = data.get("MESSAGE")
+            if msg:
+                match = secret_filter["compiled_pattern"].match(msg)
+                if match:
+                    msg = ""
+                    changed = True
+
+                    if self.reader.secret_filter_metrics is True:
+                        self.pump.stats.increase(
```
rikonen, line 740: I presume matches are generally speaking rare but if for whatever reason a lot of the log lines match the pattern this could result in a lot of statsd messages getting sent. Would seem a bit safer to just increase a counter here and have the main loop send out the value.

ryanisroot: good call, fixed

README.rst
ryanisroot: this was a great suggestion, implemented now. the concept of redaction groups is gone, and both formats of regex will work / have examples / tests

journalpump/journalpump.py
```diff
@@ -675,6 +723,16 @@ def _truncate_long_message(self, json_entry):
         }
         return json.dumps(entry, default=default_json_serialization).encode("utf8")
 
+    def _apply_secret_filters(self, data):
+        msg = data.get("MESSAGE")
+        if msg:
+            for secret_filter in self.reader.secret_filters:
+                if self.reader.secret_filter_metrics is True:
+                    self.reader.secret_filter_matches = self.reader.secret_filter_matches + 1
```
rikonen, line 731: Nothing checks here that there in fact was a match. `re.sub` doesn't directly give that info but I guess we could just see if new message is different from original message to determine that.

Also, nit: `+=` would be shorter.

README.rst
```diff
@@ -268,6 +268,17 @@ Example configuration for a single reader::
...
           }
       ],
+      "secret_filter_metrics": true,
+      "secret_filters": [
+        {
+          "pattern": "SENSITIVE",
+          "redaction_message": "[REDACTED]"
+        }
+        {
+          "pattern": "(.*?\\s?)(SECRET)(\\s?.*)",
```
rikonen, line 278: I'd use something else as an example for this because both the leading and trailing pattern match anything and this works exactly the same way as the first example. (Same for the other location that uses this same example pattern.)

README.rst
```diff
@@ -320,6 +331,45 @@ current user; and ``"OS_ROOT"`` is used to open the journal from directories rel
...
+  }
+]
+
+These examples will both match a secret pattern of ``SECRET``, whether it is at the start, middle or end of a log entry.
+It will also leave whitespace surrounding the matched group intact. The substring will be replaced inline with the configured
+redaction message. In this Example, a log entry of "This line contains a SECRET" will be translated to "This line contains a [REDACTED]"
+before being forwarding to any of the readers configured senders.
+
+
+``secret_filter_metrics`` ( default: ``false``)
```
rikonen, line 368: This is fine if as such but just wondering if being able to disable metrics output is especially useful. This should result in very low volume metric output so hard to see that being especially problematic for any use case.

journalpump/journalpump.py
```diff
@@ -495,6 +509,37 @@ def value_func(tags, f=f, args=args):  # pylint: disable=undefined-variable
...
+        for secret_filter in secret_filters:
+
+            if secret_filter.get("redaction_message") is None:
+                raise ValueError("invalid secret_filters configuration - missing field 'redaction_message'")
+
+            if secret_filter.get("pattern") is None:
+                raise ValueError("invalid secret_filters configuration - missing field 'pattern'")
+
+        # Compile / validate regex
+        for idx, _ in enumerate(secret_filters):
```
rikonen, line 535: Nit: Why ignore the actual value and access it via index instead of just having `idx, filter_info` here and `filter_info["pattern"]` later?

journalpump/journalpump.py
```diff
@@ -495,6 +509,37 @@ def value_func(tags, f=f, args=args):  # pylint: disable=undefined-variable
...
+            try:
+                secret_filters[idx]["compiled_pattern"] = re.compile(secret_filters[idx]["pattern"])
+            except:
+                raise ValueError("invalid secret_filters configuration - invalid regex") from None
```
rikonen, line 539: Is there some particular reason to hide the original error here? Could contain useful information.

rikonen (changes requested):
In general looks good to me now, some minor comments.

journalpump/journalpump.py
ryanisroot: Thanks again for taking a look Rauli. I decided to remove the "if match" for 2 reasons. 

1. the re.sub operation returns the string unchanged if there is no match, so we get the same behavior in much simpler structure, with only 1 regex search instead of 2 regex search operations. 
2. with the simpler regex pattern, re.match only works if the pattern is at the start of the string... I could swap it for something like re.search, but decided against it for reason #1. 

Also, shortened to +=  now.

README.rst
ryanisroot: I've updated the first main example to only touch on the simple example, but in the complex example description down below i'd like to keep an explanation of the second approach.

README.rst
ryanisroot: @mble requested this be behind a config flag, so i put i behind a config flag, But I'm open to either option here.

journalpump/journalpump.py
rikonen: I didn't quite follow. My point was that this will now report metric data point for every log line, no matter whether it matches or not. I suggested having something like:

```
original_msg = msg
msg = pattern.sub(...)
if original_msg != msg:
    self.reader.secret_filter_matches += 1
```

Btw. when writing that line there I realized the counter will be incremented for every sender so if you have multiple senders the count is multiplied by the number of those.

journalpump/journalpump.py
ryanisroot: fixed

journalpump/journalpump.py
ryanisroot: ahh, the first was an oversight when i was refactoring. Fixed that, but.. i hadn't noticed this code was called for every sender. I'll have to think about what to do to resolve that.

ryanisroot: I moved the _apply_filters method up to process(), so now it is run only once ( per line per reader)

README.rst
ryanisroot: I see what you mean, I updated it with a couple of examples, using simpler syntax and a more reasonable example. 
Also swapped out "redaction_message" with "replacement" too as I like your suggestion more.

journalpump/journalpump.py
```diff
@@ -495,6 +509,37 @@ def value_func(tags, f=f, args=args):  # pylint: disable=undefined-variable
 
             yield output
 
+    def _configure_secret_filter_metrics(self, config):
+        secret_filter_metrics = config.get("secret_filter_metrics")
+        if secret_filter_metrics is not None and secret_filter_metrics is True:
```
rikonen, line 514: Nit: `secret_filter_metrics is not None` has no effect

journalpump/journalpump.py
```diff
@@ -495,6 +509,37 @@ def value_func(tags, f=f, args=args):  # pylint: disable=undefined-variable
...
+    def _validate_and_build_secret_filters(self, config):
+        secret_filters = config.get("secret_filters")
+        if secret_filters is None:
+            return []
+
+        if not isinstance(secret_filters, list):
+            raise ValueError("invalid secret_filters configuration - must be a list")
+
+        for secret_filter in secret_filters:
+
```
rikonen, line 527: Nit: Extra linefeed

journalpump/journalpump.py
```diff
@@ -675,6 +724,18 @@ def _truncate_long_message(self, json_entry):
         }
         return json.dumps(entry, default=default_json_serialization).encode("utf8")
 
+    def _apply_secret_filters(self, data):
+        if self.reader.secret_filter_metrics is True:
```
rikonen, line 728: Nit: PEP-8 says this should be `if self.reader.secret_filter_metrics` but more importantly this entire check should be removed (or perhaps rather replaced with `if self.reader.secret_filters`), now if config says metrics for secret filtering shouldn't be reporting the code ends up skipping the filtering altogether instead of just the reporting part.

journalpump/journalpump.py
```diff
@@ -330,6 +334,15 @@ def get_state(self):
             "total_bytes": self.read_bytes,
         }
 
+    def inc_secret_filter_metrics(self):
```
rikonen, line 337: Nit: `inc` is a bit ambiguous here, I think `report` or something like that would be clearer.

journalpump/journalpump.py
```diff
@@ -977,6 +1038,9 @@ def run(self):
                     # Check if there is something to read in newly created reader
                     buffered_events[reader] = journal.APPEND
 
+            for reader in self.readers.values():
```
rikonen, line 1041: Should skip this if metric reporting is disabled

rikonen (changes requested):

rikonen (changes requested):
Should squash commits, otherwise looks good to me.

rikonen (approved):
Looks good to me, not sure if @mble still wanted to check so not merging just yet

PR #113, opened 2022-03-01
daemon: use setproctitle to name process journalpump

paulwouters:
This way, the process shows up as journalpump and not as one of many python processes.


fingon (changes requested):
Content looks fine, but please rebase and fix unittest failures whatever they are.

PR #114, opened 2022-03-10
rsyslog: handle epipe transparently in syslog client

mhoffm-aiven:
If sending a message through syslog fails with EPIPE because the remote socket was closed we have to have a short retry mechanism. If we are not retrying the sending immediately we could continously have a situation where the socket timed out when we want to write something into it.

### DIFF
--- a/journalpump/rsyslog.py
+++ b/journalpump/rsyslog.py
@@ -5,6 +5,7 @@
 from functools import partial
 
 import datetime
+import errno
 import socket
 import ssl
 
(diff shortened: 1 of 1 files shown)


journalpump/rsyslog.py
```diff
@@ -133,11 +149,30 @@ def close(self):
...
+                attempts += 1
+
+                if attempts > self.retry_attempts or not self._should_retry(ex=ex):
+                    raise
+
+                time.sleep(self.retry_backoff_seconds)
+
+    def _should_retry(self, *, ex):
+        if isinstance(ex, OSError):
+            return ex.errno is errno.EPIPE
```
rikonen, line 174: `is` => `==`, otherwise this emits SyntaxWarning

rikonen (changes requested):
Could also consider only catching `OSError` since we won't try retrying others but current code works just as well

journalpump/rsyslog.py
mhoffm-aiven: Thanks!

rikonen (dismissed):

safa-topal:
Nitpick: there's already some backoff retry implementation [here](https://github.com/aiven/journalpump/blob/c3b233f9b86f9af7a1b66e2d475f6954445c11ae/journalpump/senders/base.py#L115), would be great to reuse. One way of doing it: move the logic to a base class and inherit it from related senders.

mhoffm-aiven:
> Nitpick: there's already some backoff retry implementation [here](https://github.com/aiven/journalpump/blob/c3b233f9b86f9af7a1b66e2d475f6954445c11ae/journalpump/senders/base.py#L115), would be great to reuse. One way of doing it: move the logic to a base class and inherit it from related senders.

The thing is, this is only supposed to do one quick retry to reconnect on EPIPE so having exponential backoff here does not feel appropriate.

rikonen:
If we want to do something for this I'd change `retry_attempts: int` to `retry: bool` and `retry_backoff_seconds: float` to `wait_before_retry: float` to make it clear there's either one retry or not and we wait fixed amount of time before doing that retry.

journalpump/rsyslog.py
```diff
@@ -133,11 +134,26 @@ def close(self):
...
+
+                self.socket.sendall(message[:self.max_msg - 1])
+                if len(message) >= self.max_msg:
+                    self.socket.sendall(b"\n")
+
+                break
+            except Exception as ex:  # pylint: disable=broad-except
+                self.close()
+
+                if not (retry and self._should_retry(ex=ex)):
```
rikonen, line 150: Nit: imo `not retry or not self._should_retry(ex=ex)` would read easier but very subjective.

rikonen (approved):

safa-topal (approved):

PR #115, opened 2022-04-04
rsyslog: make more error codes retriable

mhoffm-aiven:
This pr aims to make the rsyslog sender more resilient to temporary networking issues

### DIFF
--- a/journalpump/rsyslog.py
+++ b/journalpump/rsyslog.py
@@ -152,7 +152,7 @@ class SyslogTcpClient:
 
     def _should_retry(self, *, ex):
         if isinstance(ex, OSError):
-            return ex.errno == errno.EPIPE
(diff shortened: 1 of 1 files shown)


rikonen:
Should squash commits. Also reason for the sleep is not obvious. What actual problem do we expect it to fix? The retry logic is for cases where connection gets unexpectedly closed and I don't know why we'd want to wait before reconnecting. There is the regular retry logic anyway that will wait for longer until it reconnects.

mhoffm-aiven:
> Should squash commits. Also reason for the sleep is not obvious. What actual problem do we expect it to fix? The retry logic is for cases where connection gets unexpectedly closed and I don't know why we'd want to wait before reconnecting. There is the regular retry logic anyway that will wait for longer until it reconnects.

done and done!

rikonen (approved):

PR #116, opened 2022-04-04
base: add heartbeat messages

mhoffm-aiven:
this pr is a suggestion of adding heartbeat messages to journalpump. related note from the datadog documentation: 

```
(Optional) Datadog cuts inactive connections after a period of inactivity. Some Rsyslog versions are not able to reconnect properly when necessary. To mitigate this issue, use time markers so the connection never stops. To achieve this, add the following 2 lines in your Rsyslog configuration:

$ModLoad immark
$MarkMessagePeriod 20
And don’t forget to restart:
```

It would be nice to have the same for journalpump

### DIFF
--- a/journalpump/senders/base.py
+++ b/journalpump/senders/base.py
@@ -81,6 +81,7 @@ class LogSender(Thread, Tagged):
         field_filter,
         stats,
         max_send_interval,
+        max_heartbeat_interval=None,
         unit_log_levels=None,
         extra_field_values=None,
         tags=None,
@@ -100,6 +101,8 @@ class LogSender(Thread, Tagged):
         self.last_send_time = time.monotonic()
         self.max_send_interval = max_send_interval
         self.max_batch_size = MAX_KAFKA_MESSAGE_SIZE
+        self.last_heartbeat_time = time.monotonic()
+        self.max_heartbeat_interval = max_heartbeat_interval
         self.batch_message_overhead = KAFKA_COMPRESSED_MESSAGE_OVERHEAD
         self.running = True
         self._sent_cursor = None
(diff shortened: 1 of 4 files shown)


journalpump/senders/rsyslog.py
```diff
@@ -53,6 +59,24 @@ def _init_rsyslog_client(self):
...
+        try:
+            self.rsyslog_client.log(
+                facility=self.config.get("heartbeat_facility", 5),
+                severity=self.config.get("heartbeat_severity", 7),
+                timestamp=datetime.datetime.utcnow().strftime("%Y-%m-%dT%H:%M:%SZ"),
+                hostname=self.config.get("heartbeat_hostname", socket.gethostname()),
+                program=self.config.get("heartbeat_program", "journalpump"),
+                msg=self.config.get("heartbeat_message", "HEARTBEAT"),
+            )
+        except Exception as ex:  # pylint: disable=broad-except
```
rikonen, line 75: Should presumably have `except RSYSLOG_CONN_ERRORS as ex:` block here as well to avoid calling `unexpected_exception` for all random network errors, which could get spammy.

rikonen (changes requested):

journalpump/senders/base.py
```diff
@@ -192,29 +196,63 @@ def request_stop(self):
...
+    def should_send_messages(self):
+        waited_for_backoff = time.monotonic() > self._backoff_dont_send_before
+        time_has_come = time.monotonic() - self.last_send_time > self.max_send_interval
+        buffer_is_full = len(self.msg_buffer) > 1000
+        return waited_for_backoff and (buffer_is_full or time_has_come)
+
+    def should_send_hearbeat(self):
+        return self.max_heartbeat_interval and time.monotonic() - self.last_heartbeat_time > self.max_heartbeat_interval
+
+    def maintenance(self):
```
rikonen, line 238: `maintenance` => `perform_maintenance`

journalpump/senders/base.py
```diff
@@ -192,29 +196,63 @@ def request_stop(self):
...
+            time.sleep(0.1)
         self.log.info("Stopping")
 
+    def should_perform_maintenance(self):
+        # Don't run maintenance operations again immediately if it just failed
+        return not self.last_maintenance_fail or time.monotonic() - self.last_maintenance_fail > 60
+
+    def should_send_messages(self):
+        waited_for_backoff = time.monotonic() > self._backoff_dont_send_before
+        time_has_come = time.monotonic() - self.last_send_time > self.max_send_interval
```
rikonen, line 231: `time_has_come` => `max_send_interval_exceeded`

journalpump/senders/base.py
```diff
@@ -192,29 +196,63 @@ def request_stop(self):
...
+        buffer_is_full = len(self.msg_buffer) > 1000
```
rikonen, line 232: `buffer_is_full` => `buffer_size_over_send_threshold`

rikonen (dismissed):
Some naming update suggestions. In general the sender base implementation looks reasonable but some of the sender subclasses need to be adjusted to take into account the fact that `backoff` doesn't block.

E.g. in `AWSCloudWatchSender` `send_messages` has

```
            self._backoff()
            self._init_logs()
```

in exception handling. That should be changed to something like

```
    def send_messages(self, *, messages, cursor):
        if self._init_logs_pending:
            self.init_logs()
            self._init_logs_pending = False
        ...
            self._backoff()
            self._init_logs_pending = True
```

Also `_init_logs` calls backoff in retry loop so that also needs to be adjusted so that it actually exits the function and we just keep track of the retry counter as instance variable. (And also the place that calls init needs to exit if init did not succeed.) 

Couple of other senders need similar changes. None of them looked to be overly complex to get working with this change.

journalpump/senders/base.py
```diff
@@ -192,29 +195,63 @@ def request_stop(self):
...
+
+            if self.should_send_hearbeat():
+                self.do_send_heartbeat()
+
+            if self.should_send_messages():
                 self.get_and_send_messages()
-            else:
-                time.sleep(0.1)
+
+            time.sleep(0.1)
```
rikonen, line 221: If we sent some messages then we may have more data to send immediately and should not sleep.

journalpump/senders/base.py
```diff
@@ -192,29 +195,63 @@ def request_stop(self):
...
+            self.stats.unexpected_exception(ex=ex, where="maintenance_operation")
+            self.last_maintenance_fail = time.monotonic()
+
+    def do_send_heartbeat(self):
+        self.log.info("Sending heartbeat message")
+        try:
+            self.send_heartbeat()
+        except Exception:  # pylint: disable=broad-except
+            self.log.exception("Problem sending heartbeat message")
+            self._backoff()
```
rikonen, line 251: backoff doesn't seem sensible for heartbeat, it's scheduled operation anyway.

journalpump/senders/rsyslog.py
```diff
@@ -53,6 +59,27 @@ def _init_rsyslog_client(self):
...
+                facility=self.config.get("heartbeat_facility", 5),
+                severity=self.config.get("heartbeat_severity", 7),
+                timestamp=datetime.datetime.utcnow().strftime("%Y-%m-%dT%H:%M:%SZ"),
+                hostname=self.config.get("heartbeat_hostname", socket.gethostname()),
+                program=self.config.get("heartbeat_program", "journalpump"),
+                msg=self.config.get("heartbeat_message", "HEARTBEAT"),
+            )
+        except RSYSLOG_CONN_ERRORS as ex:
+            self.mark_disconnected(ex)
+            raise ex
```
rikonen, line 77: Perhaps this should log warning and return instead of raising exception, otherwise base sender will print full stack trace, which might be spammy.

rikonen (changes requested):

journalpump/senders/rsyslog.py
mhoffm-aiven: i can convert the logging in the base sender to warn too, wdyt?

journalpump/senders/base.py
mhoffm-aiven: :heavy_check_mark:

journalpump/senders/base.py
mhoffm-aiven: :heavy_check_mark:

rikonen (approved):

rikonen (changes requested):
Commits should be squashed and messages updated, currently they're "wip: something"

rikonen (approved):

PR #118, opened 2022-04-19
rsyslog: remove heartbeats, ignore exceptions on close

mhoffm-aiven:
This PR removes heartbeats again since the EPIPE error handling in the
rsyslog sender was not handling errors when closing the connection.
It adds ECONNABORTED to the exceptions that should be expected and
retried once the conncetion has gone cold.


mhoffm-aiven:
closed in favor of #119 which is better factored into seperate commits

PR #120, opened 2022-04-19
Skip corrupted messages

m-novikov:
It's possible that due file system errors or some underlying issues in journald corrupted entries could be present in journal files.


journalpump/journalpump.py
```diff
@@ -377,10 +377,20 @@ def inc_line_stats(self, *, journal_lines, journal_bytes):
...
 
     def read_next(self):
+        """Read next message from journal reader."""
         if not self.journald_reader:
             return None
 
-        jobject: JournalObject = cast(JournalObject, next(self.journald_reader))
+        try:
+            jobject: JournalObject = cast(JournalObject, next(self.journald_reader))
+        except OSError as ex:
```
hackaugusto, line 386: How stable is the error message? `OSerror` is a very broad exception, it would make sense to check the exception message so that we handle only the ones related to the problem.

fingon (changes requested):
+1 on Augusto's comment. Also, if we're handling it, I'm not sure if it really classifies as unexpected exception or something we want to log using log.exception (having default handling for non-handled case should do that).

m-novikov:
Maybe it should send some metric about skipped messages to statsd?

PR #122, opened 2022-04-20
Prevent journalpump hang on exceptions

aihessu:
Prevent journalpump from getting stuck when an exception is raised. Previously, an exception which was not caught, would cause journalpump to get stuck, since the sender threads are not stopped, and they are not marked with daemon=True. Mark them as daemon, so that the main thread quitting will cause the program to quit (and get restarted by systemd).
* Tested that the "Bad message on journal read" and "OSError on stats socket write" exceptions now cause termination, not stuckness
* Tested that journalpump shutdown is still clean; the main thread will quit after websocket & kafka senders have stopped

Also catch the stats socket write error - it may happen frequently when the receiving UDP service is restarted and not available for a few seconds.

### DIFF
--- a/journalpump/senders/base.py
+++ b/journalpump/senders/base.py
@@ -86,7 +86,9 @@ class LogSender(Thread, Tagged):
         tags=None,
         msg_buffer_max_length=50000
     ):
-        Thread.__init__(self)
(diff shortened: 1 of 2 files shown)


aiven-amartin (approved):
LGTM

tkren (approved):

PR #125, opened 2022-05-09
Add OpenSearchSender

willyborankin:
Added sender for `OpenSearch` with the same config parameters as for `ElasticSearch`.
Main diff between `ElasticSearch` and `OpenSearch` senders:
- Config settings have prefix `opensearch_`
- `OpenSearchSender` checks `opensearch_url` settings during creation time 
- `OpenSearchSender` creates index mapping without type, since index type deprecated in `OpenSearch` 1.x and removed in `OpenSearch` 2.x  
- `self.request_timeout` neved used locally in `ElasticSearchSender`, changed value to `60` instead of `10` and use it in `get_requests_session`
- The code in `OpenSearchSender` was refactored a bit
- `ElasticsearchSender` and `OpenSearchSender` use same functionality as possible

This is a first part of feature I'm going to add. So the next PR will contain additional testing for index aware senders   
 
I think it makes sense to leave `ElasticSearchSender` as is for backward compatibility and mark it as `deprecated`.


mhoffm-aiven (approved):
looks good to me

rikonen:
Should presumably add some base class that both ES and OS implementations inherit to avoid duplicating code.

willyborankin:
@rikonen  Ok will try.

willyborankin:
@rikonen Added a base class both for `OpenSearchSender` and `ElasticsearchSender`. This is the first PR, in the next PR I'm going to add additional tests for senders and extract duplicate logic, which I did not touch intentionally to reduce PR size.

mhoffm-aiven (approved):
lgtm

PR #126, opened 2022-05-17
elasticsearch,opensearch: Add fields for index mapping

willyborankin:
- Add support ES 8.x and keep backward compatibility for ES 6.x
- Add index auto-mapping for fields in a message, so if JSON message has fields they will be
  mapped automatically to the index mapping. Supported python types:
    - str
    - boolean
    - int
    - float
    - dict

### DIFF
--- a/journalpump/senders/elasticsearch_opensearch_sender.py
+++ b/journalpump/senders/elasticsearch_opensearch_sender.py
@@ -1,4 +1,4 @@
-from abc import ABC, abstractmethod
+from copy import deepcopy
 from dataclasses import dataclass
 from http import HTTPStatus
 from io import BytesIO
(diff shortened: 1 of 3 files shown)


journalpump/senders/elasticsearch_opensearch_sender.py
```diff
@@ -252,47 +242,86 @@ def _create_index_and_mapping(self, index_name: str) -> None:
...
+            if isinstance(v, bool):
+                addition_fields[k] = {"type": "boolean"}
+            elif isinstance(v, float):
+                addition_fields[k] = {"type": "float"}
+            elif isinstance(v, int):
+                addition_fields[k] = {"type": "integer"}
+            elif isinstance(v, str):
+                addition_fields[k] = {"type": "text"}
+            elif isinstance(v, type(None)):  # FIXME not sure that none fields is our case but who knows
+                addition_fields[k] = {"type": "null"}
```
mhoffm-aiven, line 275: i dont think "null" is a valid type

journalpump/senders/elasticsearch_opensearch_sender.py
```diff
@@ -27,6 +43,8 @@ class Config:
 
     _DEFAULT_INDEX_PREFIX = "journalpump"
 
+    _SUPPORTED_DATA_TYPES = (type(None), bool, float, int, str)
```
mhoffm-aiven, line 46: not  used, can probably go

mhoffm-aiven (approved):
lgtm

PR #131, opened 2022-06-24
Fix index url for old ES version

willyborankin:
Fixed index url for old ES versions which needs to include
include_type_name flag


mhoffm-aiven (approved):
lgtm

PR #133, opened 2022-07-20
fix: redact sensitive information from indicies url when logging

mble:
`_indicies_url` for the `_EsOsLogSenderBase` class can contain sensitive
information, typically a password in the form of
`https://user:pass@host:port`. As a result, this information would be
logged should ES/OS not be available.

This commit redacts this using a standard pattern to remove this
password from being logged.

### DIFF
--- a/journalpump/senders/elasticsearch_opensearch_sender.py
+++ b/journalpump/senders/elasticsearch_opensearch_sender.py
@@ -10,6 +10,7 @@ from typing import Any, Dict, Set, Union
 
 import enum
 import json
+import re
 import time
 
 
@@ -71,6 +72,8 @@ class Config:
 class _EsOsLogSenderBase(LogSender):
 
     _DEFAULT_MAX_SENDER_INTERVAL = 10.0
+
+    _INDICIES_URL_REDACTION_REGEXP = r"(\w*?://[A-Za-z0-9\-._~%!$&'()*+,;=]*)(:)([A-Za-z0-9\-._~%!$&'()*+,;=]*)(@)"
 
     _ONE_HOUR_LAST_INDEX_CHECK = 3600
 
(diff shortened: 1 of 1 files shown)


journalpump/senders/elasticsearch_opensearch_sender.py
```diff
@@ -72,6 +73,8 @@ class _EsOsLogSenderBase(LogSender):
 
     _DEFAULT_MAX_SENDER_INTERVAL = 10.0
 
+    _INDICIES_URL_REDACTION_REGEXP = r"(\w*?://[A-Za-z0-9\-_]*)(:)([A-Za-z0-9\-_]*)(@)"
```
sigmaris, line 76: The user and password may be percent-encoded or have other punctuation in them, from reading https://datatracker.ietf.org/doc/html/rfc3986#section-3.2.1 I suggest this regex which matches characters from the "unreserved", "pct-encoded", and "sub-delims" elements in the username and password parts.

```suggestion
    _INDICIES_URL_REDACTION_REGEXP = r"(\w*?://[A-Za-z0-9\-._~%!$&'()*+,;=]*)(:)([A-Za-z0-9\-._~%!$&'()*+,;=]*)(@)"
```

sigmaris (approved):

journalpump/senders/elasticsearch_opensearch_sender.py
sigmaris, line 76: The user and password may be percent-encoded or have other punctuation in them, from reading https://datatracker.ietf.org/doc/html/rfc3986#section-3.2.1 I suggest this regex which matches characters from the "unreserved", "pct-encoded", and "sub-delims" elements in the username and password parts.

```suggestion
    _INDICIES_URL_REDACTION_REGEXP = r"(\w*?://[A-Za-z0-9\-._~%!$&'()*+,;=]*)(:)([A-Za-z0-9\-._~%!$&'()*+,;=]*)(@)"
```

sigmaris (changes requested):

journalpump/senders/elasticsearch_opensearch_sender.py
mble: That seems reasonable to me.

sigmaris (approved):

PR #139, opened 2022-11-30
Revert "rsyslog: mark some ssl errors as retriable"

safa-topal:
This reverts commit 99345ea684ce74c782d8a3b32d38aadccc68e171.

It turned out `SSLEOFError` was not the cause of the issue, so no need to retry for this particular error. This doesn't seem like a recoverable one anyway so it makes sense to reinitialize right away when the rsyslog client if `SSLEOFError` happens

### DIFF
--- a/journalpump/rsyslog.py
+++ b/journalpump/rsyslog.py
@@ -155,9 +155,6 @@ class SyslogTcpClient:
     def _should_retry(self, *, ex):
         if isinstance(ex, OSError):
             return ex.errno in (errno.EPIPE, errno.ECONNRESET, errno.ETIMEDOUT)
-        if isinstance(ex, ssl.SSLEOFError):
-            #  SSL connection has been terminated abruptly
-            return True
         return False
 
     def log(self, *, facility, severity, timestamp, hostname, program, pid=None, msgid=None, msg=None, sd=None):


mhoffm-aiven:
Just as a note: the connection is reinitialized when we retry.

safa-topal:
Yes, it is after one retry. I modified the description of the PR to make that point clear

mhoffm-aiven:
Are we sure that this is in general not a retriable error though?

mteiste:
You sure? E.g. there is clearly (to my understanding) an issue which results losing connections, so before accepting, what was the issue and has it been fixed?

safa-topal:
> You sure? E.g. there is clearly (to my understanding) an issue which results in losing messages, so before accepting, what was the issue and has it been fixed?

The issue with sending logs to DD was resolved after switching port to 443. For some reason our docs state another port which is not officially documented in relevant DD doc. We are soon updating the docs too to get 443 as DD port.


> Are we sure that this is in general not a retriable error though?

Yes, IMHO it's an exception saying SSL connection is quite broken. [Docs](https://docs.python.org/3/library/ssl.html#ssl.SSLEOFError) on the specific exception also advises not `to reuse the underlying transport when this error is encountered.`

mteiste:
> The issue with sending logs to DD was resolved after switching port to 443. For some reason our docs state another port which is not officially documented in relevant DD doc. We are soon updating the docs too to get 443 as DD port.

The 105xx port in the documentation used to be a valid endpoint for tcp (syslog) for US at the time the documentation was written, but yes it appears that this has now changed.

mteiste (approved):

PR #140, opened 2023-03-02
Stop (re)sending zero failure metric

fingon:
If there never was an error, sending the metric constantly is simply waste of resources.


fingon:
This saves about 2,5% of the M3 cluster memory (and some CPU and other resources elsewhere).

sjamgade (approved):

PR #143, opened 2023-03-20
remove boto3 dependency

sebinsunny:
botocore dependency is simpler for the use case of `journalpump` rather than using the boto3 wrapper. This makes it easier to use it along with other dependencies.

### DIFF
--- a/journalpump/senders/aws_cloudwatch.py
+++ b/journalpump/senders/aws_cloudwatch.py
@@ -1,7 +1,6 @@
 from .base import LogSender, SenderInitializationError
+from botocore import exceptions, session
 
-import boto3
-import botocore
 import json
 import time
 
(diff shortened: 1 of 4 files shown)


journalpump/senders/aws_cloudwatch.py
```diff
@@ -1,7 +1,7 @@
 from .base import LogSender, SenderInitializationError
 
-import boto3
 import botocore
+import botocore.session
```
lionbee, line 4: Is this import needed?

lionbee (approved):
LGTM: Overall it looks fine, the only real test will be seeing that this integration works. When you get to this on aiven-coreI may or may not be able to help you with that :)

journalpump/senders/aws_cloudwatch.py
sebinsunny: fixed the import

lionbee (approved):
LGTM

PR #144, opened 2023-03-20
ci: support python 3.11

mteiste:
the "standard" pylint action did not work for python 3.11 so replaced it with re-introduced `make pylint`.
Until full cleanup has been made only fail the build if pylint step finds Errors.
Mark the "known" import errors instead of blindly allowing all import errors unlike previously done in the commit hook implementation.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -15,22 +15,26 @@ jobs:
     strategy:
       matrix:
         # only use oldest and newest version for the lint step
-        python-version: ["3.10"]
+        python-version: ["3.9", "3.11"]
 
     steps:
 
       - id: checkout-code
         uses: actions/checkout@v2
 
+      - id: system-dependencies
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y libsystemd-dev libsnappy-dev
+
       - id: prepare-python
         uses: actions/setup-python@v2
         with:
           python-version: ${{ matrix.python-version }}
 
-      - id: dependencies
+      - id: python-dependencies
         run: |
-          sudo apt-get update
-          sudo apt-get install -y libsystemd-dev
+          python -m pip install --upgrade pip
           pip install -r requirements.txt
           pip install -r requirements.dev.txt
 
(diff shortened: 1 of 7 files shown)


.github/workflows/build.yml
```diff
@@ -43,23 +47,24 @@ jobs:
     strategy:
       max-parallel: 4
       matrix:
-        python-version: [3.8, "3.10"]
+        python-version: ["3.9", "3.10", "3.11"]
```
fingon, line 50: Could probably drop 3.10 here.

test/test_websocket.py
```diff
@@ -6,7 +6,7 @@
 import asyncio
 import json
 import logging
-import snappy
+import snappy  # pylint: disable=import-error
```
fingon, line 9: Why is this needed?

Makefile
```diff
@@ -10,6 +11,10 @@ PYLINT_DIRS = journalpump/ test/ systest/
 unittest:
 	$(PYTHON) -m pytest -vv test/
 
+.PHONY: pylint
+pylint:
+	$(PYTHON) -m pylint --rcfile .pylintrc $(PYLINT_DIRS) || exit $$(($$?&2))  # only fail on errors
```
fingon, line 16: `pylint --enable F,E` ?

Fatal, error; warning is enabled by default too I think.

I would probably define this CI-only target to be error only - running the W and I checks and ignoring their results in CI doesn't make much sense (and default pylinting presumably is done elsewhere).

fingon (changes requested):

Makefile
mteiste: because (tried) the --enable F,E has no effect on the return status

test/test_websocket.py
mteiste: most likely for the same reason the pre-commit hook mypy has `args: [--ignore-missing-imports]` and the origin pre-commit hook pylint had `[--disable=E0401]`, to be honest i did not investigate further why the snappy is not available when the pre-commit hooks are run.

fingon (approved):

PR #146, opened 2023-03-24
remove boto3 dependency

sebinsunny:
botocore dependency is simpler for the use case of journalpump rather than using the boto3 wrapper. This makes it easier to use it along with other dependencies.


lionbee (dismissed):
one last change...README.rst also needs to be updated to remove the boto3 reference

lionbee (approved):
LGTM

PR #149, opened 2023-04-14
config: handle missing match_value as described in readme

aiven-amartin:
In the docs it says we default to null for `match_value`, however we
actually throw. This makes us gracefully handle that input as described, and updates the docs to explicitly describe that behaviour.

### DIFF
--- a/journalpump/daemon.py
+++ b/journalpump/daemon.py
@@ -40,7 +40,7 @@ class ServiceDaemon:
         self.log = logging.getLogger(self.name)
         self.config_path = config_path
         self.config_file_ctime = None
-        self.config = None
+        self.config = {}
         self.require_config = require_config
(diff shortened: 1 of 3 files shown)


journalpump/daemon.py
```diff
@@ -40,7 +40,7 @@ def __init__(
         self.log = logging.getLogger(self.name)
         self.config_path = config_path
         self.config_file_ctime = None
-        self.config = None
+        self.config = {}
```
aiven-amartin, line 43: This gets initialized in the next two lines, setting here so typing checks don't complain, and its obvious that this is always a dict.

The require_config being something that can be disabled also probably doesn't make sense, as lots of code has required keys from a brief grep, but that can be a future change.

PR #150, opened 2023-04-14
add support for match_key filter at the reader level

kmichel-aiven:


journalpump/journalpump.py
```diff
@@ -675,6 +690,9 @@ def process(self) -> SingleMessageReadResult:
         if not self.pump.check_match(new_entry):
             return SingleMessageReadResult(has_more=True, bytes_read=None)
```
aiven-amartin, line 691: I guess we probably shouldn't remove this since it's existing behaviour, but I guess we should update the docs to indicate this is a top level config as well as a reader level one? And/or just log a deprecation message when top-level is configured. Can be a followup PR.

aiven-amartin (approved):
One nit around the existing behaviour, but can figure that out in another PR once we decide on the best path forward there.

PR #151, opened 2023-04-14
add missing test for match key filtering

kmichel-aiven:


alanfranz (approved):
I still think that we shouldn't use conditionals in parametrized tests (I'd prefer a MY_KEY and an expected field) but it's better to have the test than not having it.

alanfranz (approved):
lgtm

alanfranz:
I cannot merge and I cannot add other people as reviewers.

rikonen (approved):

rikonen:
Hmm, for whatever reason I also cannot merge. I have been able to merge journalpump pull requests in the past.

staaldraad (approved):

PR #153, opened 2023-04-20
update rsyslog TLS config

staaldraad:
Create a default ssl context with minimum_version set to TLSv1.2. This should be nearly universally compatible and provide the best baseline level of security/compliance

`ssl.create_default_context()` is used since we require > python3.9; 
> ssl.create_default_context - a convenience function, supported in Python 3.4 and later versions.
Even when you use these alternatives, you should ensure that a safe protocol is used. The following code illustrates how to use flags (available since Python 3.2) or the `minimum_version` field (favored since Python 3.7) to restrict the protocols accepted when creating a connection.

### DIFF
--- a/journalpump/rsyslog.py
+++ b/journalpump/rsyslog.py
@@ -97,6 +97,7 @@ class SyslogTcpClient:
             protocol = "PLAINTEXT"
         if cacerts is not None or protocol == "SSL":
             self.ssl_context = ssl.SSLContext(ssl.PROTOCOL_TLS)
+            self.ssl_context.minimum_version = ssl.TLSVersion.TLSv1_2
             self.ssl_context.verify_mode = ssl.CERT_REQUIRED
             if cacerts:
                 self.ssl_context.load_verify_locations(cacerts)


journalpump/rsyslog.py
```diff
@@ -96,7 +96,8 @@ def __init__(
         if protocol is None:
             protocol = "PLAINTEXT"
         if cacerts is not None or protocol == "SSL":
-            self.ssl_context = ssl.SSLContext(ssl.PROTOCOL_TLS)
+            self.ssl_context = ssl.create_default_context()
+            self.ssl_context.minimum_version = ssl.TLSVersion.TLSv1_2
             self.ssl_context.verify_mode = ssl.CERT_REQUIRED
```
rikonen, line 101: Looking at the source code this line is redundant (but no harm leaving it for clarity) but one thing that does change here is that `check_hostname` is set to `True`, which is sensible setting in general but we did not have that  before so not sure if something will break.

journalpump/rsyslog.py
```diff
@@ -96,7 +96,8 @@ def __init__(
...
             if cacerts:
                 self.ssl_context.load_verify_locations(cacerts)
```
rikonen, line 103: Should pass `cafile=cacerts` for `create_default_context()` instead?

journalpump/rsyslog.py
```diff
@@ -96,7 +96,8 @@ def __init__(
         if protocol is None:
             protocol = "PLAINTEXT"
         if cacerts is not None or protocol == "SSL":
-            self.ssl_context = ssl.SSLContext(ssl.PROTOCOL_TLS)
+            self.ssl_context = ssl.create_default_context()
```
rikonen, line 99: This calls `load_default_certs()`. I'm guessing that when `certfile` is present we should only accept what's in there but will that be merged with the default certs so that we accept certs that should not be accepted?

rikonen (commented):
Looks like using `create_default_context()` potentially changes the behavior quite a bit and not clear if that's a problem or not.

journalpump/rsyslog.py
staaldraad: good catch! Maybe it is better/safer then to go with 

```python
self.ssl_context = ssl.SSLContext()
self.ssl_context.options |= ssl.OP_NO_TLSv1 | ssl.OP_NO_TLSv1_1
# and (probably not necessary)
# self.ssl_context.minimum_version = ssl.TLSVersion.TLSv1_2
```

staaldraad:
> Looks like using `create_default_context()` potentially changes the behavior quite a bit and not clear if that's a problem or not.

I've changed it to create the SSLContext as before, and just set the minimum version. This doesn't seem to actually matter from 3.10 onwards, where sane minimal versions are selected (`PROTOCOL_TLS` is the default)

> Changed in version 3.10: The default cipher suites now include only secure AES and ChaCha20 ciphers with forward secrecy and security level 2. RSA and DH keys with less than 2048 bits and ECC keys with less than 224 bits are prohibited. [PROTOCOL_TLS](https://docs.python.org/3/library/ssl.html#ssl.PROTOCOL_TLS), [PROTOCOL_TLS_CLIENT](https://docs.python.org/3/library/ssl.html#ssl.PROTOCOL_TLS_CLIENT), and [PROTOCOL_TLS_SERVER](https://docs.python.org/3/library/ssl.html#ssl.PROTOCOL_TLS_SERVER) use TLS 1.2 as minimum TLS version.

journalpump/rsyslog.py
```diff
@@ -96,7 +96,8 @@ def __init__(
         if protocol is None:
             protocol = "PLAINTEXT"
         if cacerts is not None or protocol == "SSL":
-            self.ssl_context = ssl.SSLContext(ssl.PROTOCOL_TLS)
+            self.ssl_context = ssl.SSLContext()
```
rikonen, line 99: `__new__` in Python 3.11 has

```
        if protocol is None:
            warnings.warn(
                "ssl.SSLContext() without protocol argument is deprecated.",
                category=DeprecationWarning,
                stacklevel=2
            )
            protocol = PROTOCOL_TLS
```

rikonen (approved):

paulwouters:
Even better is if they would refrain from setting crypto parameters and let the system wide defaults kick in :P

PR #157, opened 2023-06-19
reader: do not emit initial 0 on always successful configurations

aiven-amartin:
#156 Attempted to fix the duplicate 0 emissions, but accidentally kept the initial 0 emission when no failures have ever occured. This corrects that test and the fix to not emit the initial 0.


mhoffm-aiven (approved):
lgtm

PR #161, opened 2023-11-15
Logging improvements for multiple senders and consistency fixes

aindriu-aiven:
I have added some additional logging to improve traceability when working with multiple senders. Now the name of the sender in the journalpump.json is appended to Initializtion logging messages to identify which senders are having issues and which are performing correctly.

e.g in the default journalpump.json "elastic1" and "kafka1" are now easily identified despite both being in the "host1" reader config.

I also added some consistency updates to the log files to make sure they all followed the same pattern.


jlprat (approved):
LGTM Thanks @aindriu-aiven

PR #162, opened 2023-11-20
Updating version and NEWS with last updates since version 2.4.3

aindriu-aiven:
Creating a release version of journal pump.
There have been tags created for 2.4.4 & 2.4.5 but they have not updated the version in __init__.py 
This will upgrade the release version and can be tagged as a release of journalpump.
Also the NEWS has been updated to include all changes since tag 2.4.3


jlprat (approved):
LGTM

PR #164, opened 2024-02-02
misc: use re2 if configured

mhoffm-aiven:
Make it possible to use re2 as a non-backtracking engine to evaluate searches. We make it opt-in, configurable using environment since otherwise it would constitute a breaking change.


aiven-amartin (approved):
LGTM, @mhoffm-aiven if you can update the description a bit with context, will hold on a second reviewer before merging. [SRE-8038]

aiven-amartin (approved):

exg (approved):

PR #165, opened 2024-02-02
Add a threshold for search if want to emit a metric

Aiqin-Aiven:
in case some bad regex search defined,
it would be easier to identify the offender

The threshold is defined as threshold_for_metric_emit with default value 10 seconds.

The metric would look like this:
image
with regex tag on it.

### DIFF
--- a/journalpump/journalpump.py
+++ b/journalpump/journalpump.py
@@ -158,6 +158,7 @@ class JournalReader(Tagged):
         self.secret_filter_metrics = self._configure_secret_filter_metrics(config)
         self.secret_filter_metric_last_send = time.monotonic()
         self._is_ready = True
+        self.threshold_for_metric_emit = self._configure_threshold_for_metric_emit(config)
 
     def invalidate(self) -> None:
         """
@@ -551,6 +552,9 @@ class JournalReader(Tagged):
 
         return secret_filters
 
+    def _configure_threshold_for_metric_emit(self, config):
+        return int(config.get("threshold_for_metric_emit", 10))
+
     def perform_searches(self, jobject):
         entry = jobject.entry
         results = {}
@@ -580,7 +584,16 @@ class JournalReader(Tagged):
                             break
                         byte_fields[field] = line
 
+                start_time = time.perf_counter()
                 match = regex.search(line)
+                regex_search_duration = time.perf_counter() - start_time
+                if regex_search_duration > self.threshold_for_metric_emit:
+                    self.stats.gauge(
+                        metric="journal.perform_search_regex_duration",
+                        value=regex_search_duration,
+                        tags=self.make_tags({"regex": regex.pattern}),
+                    )
+                    self.log.info("Slow regex search: %s for duration %s seconds", regex, regex_search_duration)
                 if not match:
                     all_match = False
                     break
--- a/README.rst
+++ b/README.rst
@@ -279,6 +279,7 @@ Example configuration for a single reader::
           "pattern": "SENSITIVE",
           "replacement": "[REDACTED]"
         }],
+      "threshold_for_metric_emit": 10
       "tags": {
           "type": "container"
       }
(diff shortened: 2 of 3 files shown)


mhoffm-aiven:
What if the search never finishes though? I think this is better served by converting to using re2!

Aiqin-Aiven:
> ishes though? I think this is better served by converting to using re2!
We had a discussion in the team and journalpump is going to be replaced by vector, so we stick to the plan to get some metrics out.

journalpump/journalpump.py
```diff
@@ -551,6 +552,18 @@ def _validate_and_build_secret_filters(self, config):
...
+    def _configure_threshold_for_metric_emit(self, config):
+        threshold_for_metric_emit = config.get("threshold_for_metric_emit")
+        if threshold_for_metric_emit is not None:
+            try:
+                threshold = int(threshold_for_metric_emit)
+            except ValueError:
+                raise ValueError("Invalid value for threshold_for_metric_emit. Must be an integer.")
+        else:
+            # Use a default value if the "threshold_for_metric_emit" is not present
+            threshold = int(10)
```
allhailwesttexas, line 564: nit: this could be `threshold = 10`.

journalpump/journalpump.py
```diff
@@ -551,6 +552,18 @@ def _validate_and_build_secret_filters(self, config):
 
         return secret_filters
 
+    def _configure_threshold_for_metric_emit(self, config):
+        threshold_for_metric_emit = config.get("threshold_for_metric_emit")
+        if threshold_for_metric_emit is not None:
+            try:
+                threshold = int(threshold_for_metric_emit)
+            except ValueError:
+                raise ValueError("Invalid value for threshold_for_metric_emit. Must be an integer.")
```
allhailwesttexas, line 561: Is this error message much clearer that what is given by python by default? Since we are reraising the same exception. The standard one is `ValueError: invalid literal for int() with base 10: '<input_value>'` which in some ways is clearer since it echoes the input value. It also accepts things like `int('25')` so it is not technically correct that it needs to be an integer.

allhailwesttexas (dismissed):
Some small comments. Also, is there any test coverage for this change?

journalpump/journalpump.py
```diff
@@ -551,6 +552,18 @@ def _validate_and_build_secret_filters(self, config):
 
         return secret_filters
 
+    def _configure_threshold_for_metric_emit(self, config):
+        threshold_for_metric_emit = config.get("threshold_for_metric_emit")
+        if threshold_for_metric_emit is not None:
+            try:
```
allhailwesttexas, line 558: Isn't this the same as having no try/except block and just doing `threshold = int(threshold_for_metric_emit)`?

Aiqin-Aiven: yep, I was just supposing some non-sense input like a string like "A", "B" was put as the config.

Aiqin-Aiven: Oh, yes, I got your point, I didn't add much extra info to the raise, so no need to do a try except here

journalpump/journalpump.py
```diff
@@ -551,6 +552,15 @@ def _validate_and_build_secret_filters(self, config):
...
         return secret_filters
 
+    def _configure_threshold_for_metric_emit(self, config):
+        threshold_for_metric_emit = config.get("threshold_for_metric_emit")
+        if threshold_for_metric_emit is not None:
+            threshold = int(threshold_for_metric_emit)
+        else:
+            # Use a default value if the "threshold_for_metric_emit" is not present
+            threshold = int(10)
+        return threshold
```
RommelLayco, line 562: nit
```suggestion
        return int(config.get("threshold_for_metric_emit", 10))
```

RommelLayco (dismissed):
i don't see the test you added

Aiqin-Aiven:
> i don't see the test you added

hmm, it was there at commit: [7d0e907](https://github.com/Aiven-Open/journalpump/commit/7d0e9076a91ad6a0339896795697b919300d4ccf), I did something wrong yesterday probably

RommelLayco (approved):
lgtm

PR #167, opened 2024-03-15
fix: resume from reader cursor when no senders configured

ngilles-aiven:
In cases where a reader was configured with no senders (eg. only configured with searches), then in case of reloads/restarts the resume cursor was assumed `None`, and the default 'initial_position' of 'head' is fallen back to. This would end up causing all matching searchg to resend metric events immediately.


aiven-amartin (approved):
Nice, LGTM! Would be good to tackle the TODO around oldest sender, but that can come later.

PR #170, opened 2024-05-24
senders: check tags for None keys/values

tkren:
Only include tags that have no None-valued keys/values.


giacomo-alzetta-aiven (changes requested):
Should add some test to verify the behavior and avoid regressions in the future.

bugant (approved):

PR #171, opened 2024-06-14
senders: Mitigate issue with GCP logging quota

italomg:
https://cloud.google.com/logging/docs/reference/v2/rest/v2/entries/write
https://cloud.google.com/logging/quotas

# WHAT

This PR is doing two things:

1. Not retrying to send logs that return with HTTP status code 4xx. Instead those logs are simply skipped.
2. Filtering out any `LogEntry` that is bigger than 256K

### DIFF
--- a/journalpump/senders/google_cloud_logging.py
+++ b/journalpump/senders/google_cloud_logging.py
@@ -23,6 +23,15 @@ class GoogleCloudLoggingSender(LogSender):
         0: "EMERGENCY",
     }
 
+    # A bit on the safe side, not exactly 256KB but this
+    # is an approximation anyway
+    # according to https://cloud.google.com/logging/quotas
+    _LOG_ENTRY_QUOTA = 250 * 1024
+
+    # Somewhat arbitrary maximum message size choosen, this gives a 56K
+    # headroom for the other fields in the LogEntry
+    _MAX_MESSAGE_SIZE = 200 * 1024
+
     def __init__(self, *, config, googleapiclient_request_builder=None, **kwargs):
         super().__init__(config=config, max_send_interval=config.get("max_send_interval", 0.3), **kwargs)
         credentials = None
@@ -62,6 +71,18 @@ class GoogleCloudLoggingSender(LogSender):
         for message in messages:
             msg_str = message.decode("utf8")
             msg = json.loads(msg_str)
+
+            # This might not measure exactly 256K but should be a good enough approximation to handle this error.
+            # We try truncating the message if it isn't possible then it is skip.
+            if len(message) > self._LOG_ENTRY_QUOTA:
+                DEFAULT_MESSAGE = "Log entry can't be logged because its size is greater than GCP logging quota of 256K"
+                if "MESSAGE" in msg:
+                    msg["MESSAGE"] = f'{msg["MESSAGE"][:self._MAX_MESSAGE_SIZE]}[MESSAGE TRUNCATED]'
+                    messsage_size = len(json.dumps(msg, ensure_ascii=False).encode("utf-8"))
+                    if messsage_size > self._LOG_ENTRY_QUOTA:
+                        msg = {"MESSAGE": DEFAULT_MESSAGE}
+                else:
+                    msg = {"MESSAGE": DEFAULT_MESSAGE}
             timestamp = msg.pop("timestamp", None)
             journald_priority = msg.pop("PRIORITY", None)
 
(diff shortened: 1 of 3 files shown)


journalpump/senders/google_cloud_logging.py
```diff
@@ -83,6 +83,18 @@ def send_messages(self, *, messages, cursor):
             self._logs.entries().write(body=body).execute()  # pylint: disable=no-member
             self.mark_sent(messages=messages, cursor=cursor)
             return True
+        except HttpError as err:
+            # Retrying a 400 will probably result in the same error
+            # so in this case we skip this LogEntry but still raise
+            # the exception for keeping visibility of the error and
+            # metrics
+            if isinstance(err, HttpError) and err.status_code >= 400 and err.status_code < 500:
+                self.mark_sent(messages=messages, cursor=cursor)
```
dogukancagatay, line 104: I aggree that 4XX errors seem unrecoverable, but there is the possibility that the authentication or authorization could be recovered, or updated after some time. Marking the messages sent would move the sender's cursor, and the sender won't try to send those messages again.

Isn't it enough to mark it disconnected (pausing the sender) and not try to connect until some action is taken? Or we can considerably increase the size of the backoff mechanism's retry intervals, and after some number of attempts, we can just stop the sender.

journalpump/senders/google_cloud_logging.py
```diff
@@ -61,7 +66,14 @@ def send_messages(self, *, messages, cursor):
 
         for message in messages:
             msg_str = message.decode("utf8")
-            msg = json.loads(msg_str)
+            # This might not measure exactly 256K but should be
+            # a good enough approximation to handle this error.
+            # In anycase, we are handling this error in the
+            # exception handler below also.
+            if len(message) > GoogleCloudLoggingSender._LOG_ENTRY_QUOTA:
+                msg = { "MESSAGE": "Log entry cannot be logged because its size is greater than GCP logging quota of 256K" }
```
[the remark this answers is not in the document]
italomg: the message is simply decoded as a JSON. AFAIK the scheme does not have required fields, so there is not guarantee about how we can/should parse the JSON.
That said, we can do a "best effort" and split the "MESSAGE" field only.
I did it this way because it was an acceptable trade of when I talked to Julien Clarysse

journalpump/senders/google_cloud_logging.py
italomg: > Isn't it enough to mark it disconnected (pausing the sender) and not try to connect until some action is taken?
In the case of a big message probably not, but, as you mentioned, in case of authentication/authorization error that can certainly change after a configuration change.

> Or we can considerably increase the size of the backoff mechanism's retry intervals, and after some number of attempts, we can just stop the sender.

I thought of having a maximum number of attempts, but the logs are transmitted in batches and the operation is atomic, so it could be hard to know which one(s) of the logs is causing problems.

italomg: I am not sure I will go forward with this approach of handling 4xx though, I have to think about it a bit more

journalpump/senders/google_cloud_logging.py
```diff
@@ -62,6 +71,18 @@ def send_messages(self, *, messages, cursor):
         for message in messages:
             msg_str = message.decode("utf8")
             msg = json.loads(msg_str)
+
+            # This might not measure exactly 256K but should be a good enough approximation to handle this error.
+            # We try truncating the message if it isn't possible then it is skip.
+            if len(message) > GoogleCloudLoggingSender._LOG_ENTRY_QUOTA:
```
dogukancagatay, line 77: Accessing a private/protected members from the public interface is not generally desirable.

Make the `LOG_ENTRY_QUOTA` member public (w/o `_`).
```suggestion
            if len(message) > GoogleCloudLoggingSender.LOG_ENTRY_QUOTA:
```
Or, access it with `self`.
```suggestion
            if len(message) > self._LOG_ENTRY_QUOTA:
```

dogukancagatay (commented):
Great work, just a few suggestions.

journalpump/senders/google_cloud_logging.py
```diff
@@ -62,6 +71,18 @@ def send_messages(self, *, messages, cursor):
...
+                DEFAULT_MESSAGE = "Log entry can't be logged because its size is greater than GCP logging quota of 256K"
+                if "MESSAGE" in msg:
+                    msg["MESSAGE"] = f'{msg["MESSAGE"][:GoogleCloudLoggingSender._MAX_MESSAGE_SIZE]}[MESSAGE TRUNCATED]'
+                    messsage_size = len(json.dumps(msg))
```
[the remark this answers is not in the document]
italomg: agreed! That's a nice addition

dogukancagatay (approved):
LGTM

HeikkiLeskinen (approved):

PR #173, opened 2025-05-07
senders: resolve and use a random address on websockets connect

hnousiainen:
Resolve websocket remote address and pick a random one as an explicit host to connect. The combination of asyncio websockets and asyncio create_connect() have some overlapping timeouts that lead to connection attempts to run through the whole list of targets returned by getaddrinfo(). This breaks connectivity in some dual stack IPv6/IPv4 cases, when either one of the address family has e.g. routing problems.

We can fairly safely pick one at random here; we have retry logic built around the connections anyway that eventually lead to connected state.


journalpump/senders/websocket.py
```diff
@@ -176,10 +179,27 @@ async def websocket_connect_coro(self):
...
+        else:
+            # Resolve hostname and pick one address at random
+            # Websockets.connect() and underlying asyncio.loop.create_connection() have some overlapping timeouts
+            # that lead to little bit difficulties with working through all addresses returned by getaddrinfo.
+            # We pick one at random here, and rely our own outer loops to handle retries.
+            present_addrs = [
+                gai_result[4][0]
+                for gai_result in await self.websocket_loop.getaddrinfo(
+                    url_parsed.hostname, 0, type=socket.SOCK_STREAM, proto=socket.IPPROTO_TCP
+                )
```
ngilles-aiven, line 191: I feel the [4][0] look a bit "magic" and/or obscure even though I suppose these are "well known":

Could the readability of this be improved with destructuring assignment, such as
 
```suggestion
                sockaddr[0] # Get IP address only
                for _, _, _, _, sockaddr in await self.websocket_loop.getaddrinfo(
                    url_parsed.hostname, 0, type=socket.SOCK_STREAM, proto=socket.IPPROTO_TCP
                )
```

or even:
```suggestion
                ip_address
                for _, _, _, _, (ip_address, *_) in await self.websocket_loop.getaddrinfo(
                    url_parsed.hostname, 0, type=socket.SOCK_STREAM, proto=socket.IPPROTO_TCP
                )
```

aihessu: The improvement works fine as long as they don't decide to add a 6th element to the tuple. Maybe an extra *_ would avoid that risk.

journalpump/senders/websocket.py
```diff
@@ -176,10 +179,29 @@ async def websocket_connect_coro(self):
...
+                preferred_host = url_parsed.hostname
 
         ws_compr = None if self.websocket_compression == WebsocketCompression.none else str(self.websocket_compression)
-        return await websockets.connect(  # pylint:disable=no-member
+        return await websockets.connect(
             self.websocket_uri,
+            host=preferred_host,
             ssl=ssl_context,
             compression=ws_compr,
             extra_headers=headers,
```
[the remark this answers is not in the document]
hnousiainen: Yup, I opened a can of worms with the modification. Fedora still ships 12.0 while PyPI carries 15.0.1. I added a compat layer on the keyword args.

journalpump/senders/websocket.py
hnousiainen: I adjusted the code with a comment and the first proposed form. getaddrinfo() is used quite widely, changes to the return types would be quite disruptive so I think it's good with the parsing of the 5-tuple.

PR #191, opened 2026-03-10
setup: add missing dependencies and version pins to install_requires

dogukancagatay:
Add python-snappy and google-re2 which are used in production code
but were only listed in requirements.txt. Sync version constraints
(kafka-python<2.0.4, websockets<14.0) from requirements.txt to
keep both files consistent.


ngilles-aiven (approved):
Maybe we shouldn't have multiple source for the same information 🤔

ngilles-aiven (changes requested):
What about `https://github.com/systemd/python-systemd/zipball/master` which is also in the `requirements.txt` is that not actually required?

ngilles-aiven (approved):

PR #193, opened 2026-05-06
rsyslog: fix octet-counted framing on oversize messages

dogukancagatay:
When `octet_counted_framing` is enabled and the encoded message
exceeds `max_message_size`, the previous code advertised the full
body length in the frame prefix but only sent a truncated body
(plus a stray newline). RFC 6587 octet-counted receivers then
consumed bytes from the following frame to make up the count,
desyncing the entire connection.

Truncate the body before computing the prefix length, drop the
non-transparent newline in octet-counted mode, and document that
`max_message_size` bounds the body bytes in both modes.


jclarysse (approved):
LGTM 👍

PR #194, opened 2026-05-11
journalpump: inject __SEQNUM into journal entries

ezotrank:
To get the log sequence number, to use it after in sort order.

Parse the sequence number from the i= field of the journald cursor string and include it as an integer under __SEQNUM in each entry returned by PumpReader.get_next(). This allows downstream consumers to use it as a secondary sort key for log lines sharing the same microsecond timestamp (e.g. MySQL crash stack traces).

No direct API exists on the systemd.journal.Reader C extension for sequence numbers, so the cursor string is the only available source, consistent with how __REALTIME_TIMESTAMP is injected via _get_realtime().

### DIFF
--- a/journalpump/journalpump.py
+++ b/journalpump/journalpump.py
@@ -102,6 +102,17 @@ class PumpReader(journal.Reader):
 
         return output
 
+    @staticmethod
+    def _parse_seqnum_from_cursor(cursor) -> int | None:
+        # cursor format: s=...;i=<hex_seqnum>;b=...;m=...;t=...;x=...
+        for part in cursor.split(";"):
+            if part.startswith("i="):
+                try:
+                    return int(part[2:], 16)
+                except ValueError:
+                    return None
+        return None
+
     def get_next(self, skip=1):
         # pylint: disable=no-member, protected-access
         """Private get_next implementation that doesn't store the cursor since we don't want it"""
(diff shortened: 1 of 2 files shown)


journalpump/journalpump.py
```diff
@@ -102,14 +102,26 @@ def convert_entry(self, entry):
 
         return output
 
+    @staticmethod
+    def _parse_seqnum_from_cursor(cursor):
+        # cursor format: s=...;i=<hex_seqnum>;b=...;m=...;t=...;x=...
+        for part in cursor.split(";"):
+            if part.startswith("i="):
+                return int(part[2:], 16)
```
dogukancagatay: We should handle the `ValueError` `int()` may possibly throw.

test/test_journalpump.py
```diff
@@ -261,6 +262,25 @@ def test_journalpump_init(tmpdir):  # pylint: disable=too-many-statements
             assert isinstance(s, GoogleCloudLoggingSender)
 
 
+def test_pump_reader_get_next_includes_seqnum():
+    reader = PumpReader()
+    cursor = "s=abc;i=1f4;b=def;m=123;t=456;x=789"
```
dogukancagatay, line 267: We should add tests for the edge cases. e.g. no seqnum field, empty seqnum, etc.

test/test_journalpump.py
```diff
@@ -261,6 +262,25 @@ def test_journalpump_init(tmpdir):  # pylint: disable=too-many-statements
...
+        mock.patch.object(journal.Reader, "_next", return_value=True),
+        mock.patch.object(journal.Reader, "_get_all", return_value=entry_data),
+        mock.patch.object(reader, "_get_realtime", return_value=1000000),
+        mock.patch.object(reader, "_get_cursor", return_value=cursor),
+    ):
+        result = reader.get_next()
+
+    assert result is not None
+    assert "__SEQNUM" in result.entry
+    assert result.entry["__SEQNUM"] == 0x1F4
```
dogukancagatay: nit: could use named constant to be more descriptive.

journalpump/journalpump.py
```diff
@@ -102,14 +102,26 @@ def convert_entry(self, entry):
 
         return output
 
+    @staticmethod
+    def _parse_seqnum_from_cursor(cursor):
```
dogukancagatay: Missing return type hint.
```suggestion
    def _parse_seqnum_from_cursor(cursor) -> int | None:
```

dogukancagatay (approved):
LGTM

PR #195, opened 2026-05-15
Bump kafka-python dependency to support 2.3.1

tvainika:
Update the version constraint from <2.0.4 to support latest 2.3.1 release.

kafka-python 2.2.0+ removed errors.RETRY_ERROR_TYPES in favor of per-class retriable attributes. Add _get_retriable_kafka_errors() helper that falls back to introspecting error classes when the legacy constant is unavailable, maintaining compatibility with both old and new versions.

### DIFF
--- a/journalpump/senders/kafka_sender.py
+++ b/journalpump/senders/kafka_sender.py
@@ -2,6 +2,7 @@ from .base import LogSender
 from kafka import errors, KafkaAdminClient, KafkaProducer
 from kafka.admin import NewTopic
 
+import inspect
 import logging
 import socket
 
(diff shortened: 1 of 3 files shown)


journalpump/senders/kafka_sender.py
```diff
@@ -15,7 +16,23 @@
 except ImportError:
     zstd = None
 
-KAFKA_CONN_ERRORS = tuple(errors.RETRY_ERROR_TYPES) + (
+
+def _get_retriable_kafka_errors():
+    """Build the set of retriable Kafka error types.
+
+    kafka-python < 2.1 provides errors.RETRY_ERROR_TYPES.
```
ezotrank, line 23: Let's support only a new kafka-python version?

tvainika: My plan was that I can upgrade journalpump first, and then continue with kafka-python packaging again. This makes it possible to break this into pieces instead of huge upgrade at once. If there is other dependencies of kafka-python than journalpump, that would be again a problem.

ezotrank (approved):
lgtm

PR #198, opened 2026-06-09
rsyslog: add opt-in escape_newlines option

dogukancagatay:
Multi-line records (e.g. Java stack traces) sent to collectors, which
support only newline-delimited log intakes, are truncated to the first
line, because the intake treats each physical newline as a record
boundary.

Add `escape_newlines` (default false) to SyslogTcpClient. When
enabled, embedded `\r` and `\n` in the `MESSAGE` field are replaced with
literal `\r` and `\n` before formatting (`\\r`, and `\\n`), so the
framed record is one physical line terminated by a single trailing
newline.

### DIFF
--- a/journalpump/rsyslog.py
+++ b/journalpump/rsyslog.py
@@ -87,6 +87,7 @@ class SyslogTcpClient:
         certfile=None,
         log_format=None,
         octet_counted_framing=None,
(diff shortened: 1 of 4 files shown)


ezotrank (approved):
lgtm

PR #199, opened 2026-06-09
journalpump: Correct seek_tail reader positioning and harden rsyslog systests

dogukancagatay:
The journal reader's seek_tail code path called `sd_journal_next()`
immediately after `sd_journal_seek_tail()`, violating the API contract.
Per the systemd man page, `sd_journal_previous()` must be called first
to land on a real entry before stepping past it. Without this, the
inotify fd was never armed and new journal entries were silently missed.

ref: https://www.man7.org/linux/man-pages/man3/sd_journal_next.3.html

The rsyslog systests had two issues that masked this bug and caused
failures on machines with a large journal. Readers with no
`initial_position` defaulted to head, flooding the sender buffer with
old entries before test messages could arrive. The oversize-frame test
used `max_message_size` 80, which truncated message bodies before the
identifier on long-hostname machines. Both are fixed in the test config.

### DIFF
--- a/journalpump/journalpump.py
+++ b/journalpump/journalpump.py
@@ -471,6 +471,19 @@ class JournalReader(Tagged):
             self.read_next()
         elif self.initial_position == "tail":
             self.journald_reader.seek_tail()
+            # sd_journal_next() is not allowed immediately after sd_journal_seek_tail();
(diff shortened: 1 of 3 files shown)


ezotrank (approved):
lgtm

PR #200, opened 2026-06-09
journalpump: replace deprecated datetime.utcnow/utcfromtimestamp calls

dogukancagatay:
Python 3.12+ emits DeprecationWarnings for these APIs, and they showed
up in our 3.14 CI runs. We swapped them for timezone-aware UTC
equivalents:

* `utcnow()` -> `now(timezone.utc)`
* `utcfromtimestamp()` → `fromtimestamp(..., timezone.utc)`

Naive UTC output is preserved via `.replace(tzinfo=None)` so downstream
timestamp formatting stays the same.


watson28 (approved):

ezotrank (approved):
lgtm

